PR #5, opened 2019-06-17
Docs and cleanup.

JnBrymn-EB:
* I also added something to notify me when a user receives a greeting.
* I also created a dev-requirements.txt


nchouard:
Doc strings, I like it!

PR #6, opened 2019-06-21
simplified greeting bot

JnBrymn-EB:
I removed a lot of code that we won't need to use if we fall back to Google Forms for now.

The resulting on-boarding message looks like this: 
image

And this is the form that we use https://docs.google.com/forms/d/e/1FAIpQLSddtT25Jw3PZt2hCxLJupqONTrmlNz4d1nXYJqk-XxPExLRhA/viewform (fill it out!)

I do have a little frustration that we have to ask for the users' email addresses because there's no way to link the form to the user. I was hoping that I could pass default answers in the URL but this doesn't seem to be the case. Still I'm leaning towards a Google Form for now so that we can tie this off and move on to the next thing. We'll revisit when we build out the user model.


JnBrymn:
I'm abandoning this since there's no easy way to link a user to a google form submission except for user-supplied email address.

PR #11, opened 2019-08-14
Added landing page to django app

nchouard:


penny_u_django/home/urls.py
```diff
@@ -0,0 +1,7 @@
+from django.urls import path
+
+from . import views
+
+urlpatterns = [
+    path('', views.index, name='index')
+]
```
JnBrymn: fix this!

PR #16, opened 2019-08-23
sticking db.sqlite3 and messages.log in mounted_files

### LINKED ISSUE #14: Set up Postgres and Dockerize
* Set up Postgres so that it works from a Docker container but it's data is persisted in a Docker Volume
* Fix logging so that log files don't disappear when Docker container is stopped. (`/var/log`?)

JnBrymn:
Works when run with:

`docker run -p 80:80 --mount type=bind,source="$(pwd)"/mounted_files,target=/usr/src/app/mounted_files -e SLACK_API_KEY=$(cat ../../slack_api_key.txt) penny_u_django`

closes #14

### DIFF
--- a/penny_u_django/bot/tests/processors/test_greeting.py
+++ b/penny_u_django/bot/tests/processors/test_greeting.py
@@ -2,7 +2,6 @@ import pytest
 
 from bot.models import User
 from bot.processors.base import Event
-import bot.processors.greeting
 from bot.processors.greeting import GreetingBotModule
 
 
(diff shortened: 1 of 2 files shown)


JnBrymn:
@Raab70 @sburns take a look at what I'm attempting here and tell me if there's a normal standardized way to do it. This is lightweight, so I kinda like it. Though, should I be creating a `docker volume`?

Raab70:
> @Raab70 @sburns take a look at what I'm attempting here and tell me if there's a normal standardized way to do it. This is lightweight, so I kinda like it. Though, should I be creating a `docker volume`?
@JnBrymn 
Yeah I've never seen this done. Typically if you want some persistence you'll use a docker volume mounted to the host filesystem. So you can create a directory on the host like `/sqlitedb/` and mount that into your container `-v /sqlitedb/:/data/`. However at that point I start to seriously question if sqlite is sufficient. IMHO it's really only good for local development.

JnBrymn:
Everyone made fun of me for this :P 

Closing

PR #41, opened 2019-09-29
Created tunnelling passthrough

JnBrymn:
Steps:

1. deploy app to https://yourapp.herokuapp.com
2. visit https://yourapp.herokuapp.com/bot and see default response "at least something worked"
3. modify the bot index endpoint to say "something else"
4. set up local tunnel (ngrock, etc.)
5. visit http://yourapp.herokuapp.com/forward?host=yourngrockhost&port=8001  (only use port if needed)
6. now visit https://yourapp.herokuapp.com/bot and see "something else"

Based upon this you can automatically forward all traffic over the tunnel.

To cancel just visit http://yourapp.herokuapp.com/forward

### DIFF
--- a/penny_university/settings.py
+++ b/penny_university/settings.py
@@ -24,7 +24,7 @@ BASE_DIR = os.path.dirname(os.path.dirname(os.path.abspath(__file__)))
 SECRET_KEY = '7e@1*$kub8yxjd&pkej#+0k+e9omlz!31$zuq(4$rglm4i(hp1'
 
 # SECURITY WARNING: don't run with debug turned on in production!
-DEBUG = True
+DEBUG = os.environ.get('PENNY_DEBUG', False)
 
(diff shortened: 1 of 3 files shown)


penny_university/settings.py
```diff
@@ -24,7 +24,7 @@
 SECRET_KEY = '7e@1*$kub8yxjd&pkej#+0k+e9omlz!31$zuq(4$rglm4i(hp1'
 
 # SECURITY WARNING: don't run with debug turned on in production!
-DEBUG = True
+DEBUG = os.environ.get('PENNY_DEBUG', '').lower() == "true"
```
nchouard: 🤕 oops!

nchouard: Good catch, but for this why can't you just do `os.eviron.get('PENNY_DEBUG', False)`?

Issue #38, opened 2019-09-09
Make Penny health slack command


JnBrymn:
`/penny check` - returns printout of health

* what should we check?
* make sure this is available to admins only
* this probably belongs in a new module separate from other user commands

PR #42, opened 2019-10-01
Pennychat

JnBrymn:
closes https://github.com/penny-university/penny_university/issues/38


bot/processors/pennychat.py
```diff
@@ -0,0 +1,289 @@
...
+        self.slack = slack
+
+    @classmethod
+    def create_penny_chat(cls, slack, event):
+        response = slack.chat_postMessage(channel=event['channel_id'],
+                                          user=event['user_id'],
+                                          blocks=create_penny_chat_blocks()
+                                          )
+
+        PennyChat.objects.create(template_channel=response.data['channel'], template_timestamp=response.data['ts'])
```
JnBrymn: how do we get this object once it's created? I see a call below to `PennyChat.objects.get(template_timestamp=event['container']['message_ts'])` but will that be a dependable way to retrieve the correct object? What about when we're popular and multiple penny chats are scheduled every second? :D

bot/processors/pennychat.py
```diff
@@ -0,0 +1,289 @@
...
+class PennyChatBotModule(BotModule):
+    def __init__(self, slack):
+        self.slack = slack
+
+    @classmethod
+    def create_penny_chat(cls, slack, event):
+        response = slack.chat_postMessage(channel=event['channel_id'],
+                                          user=event['user_id'],
+                                          blocks=create_penny_chat_blocks()
+                                          )
```
JnBrymn: The docs I've been reading indicate that you can open up a modal directly when triggered by a command. E.g. we can skip ☝️ this message from being sent and just go ahead and open the modal

bot/processors/pennychat.py
```diff
@@ -0,0 +1,289 @@
...
+    def open_details_view(self, event):
+        penny_chat = PennyChat.objects.get(template_timestamp=event['container']['message_ts'])
+        modal = penny_chat_details_modal(penny_chat)
+        response = self.slack.views_open(view=modal, trigger_id=event['trigger_id'])
+        penny_chat.view = response.data['view']['id']
+        penny_chat.save()
+
+    @is_event_type('view_submission')
+    @has_callback_id('penny_chat_details')
+    def submit_details(self, event):
```
JnBrymn: Would it simplify the implementation if we just created the `PennyChat` object with the data from the submitted view? If the user abandons the flow half way through, we will have a partially created penny chat object.

bot/models.py
```diff
@@ -15,3 +15,14 @@ class User(models.Model):
...
+
+class PennyChat(models.Model):
+    title = models.TextField()
+    description = models.TextField()
+    date = models.DateTimeField()
+    invitees = models.TextField()
+    channels = models.TextField()
+    view = models.TextField()
+    template_timestamp = models.TextField()
+    template_channel = models.TextField()
```
JnBrymn: Let's talk about adding a `status`. We did this for events at Eventbrite. The status was a number with a meaning - like 10=planning, 20=scheduled, 30=completed, 40=canceled - something like that.

bot/processors/pennychat.py
```diff
@@ -0,0 +1,402 @@
...
+
+    @is_block_interaction_event
+    @is_action_id('penny_chat_date')
+    def date_select(self, event):
+        date = event['actions'][0]['selected_date']
+        penny_chat = PennyChat.objects.get(view=event['view']['id'])
+        tz = timezone(penny_chat.user_tz)
+        time = str(penny_chat.date.astimezone(tz).time()) if penny_chat.date else '00:00:00'
+        penny_chat.date = tz.localize(datetime.strptime(date + ' ' + time, '%Y-%m-%d %H:%M:%S'))
+        penny_chat.save()
```
JnBrymn: nit pick - put these in the same order as they are in the modal

bot/processors/pennychat.py
```diff
@@ -0,0 +1,402 @@
...
+def shared_message_template(penny_chat):
+    timestamp = int(penny_chat.date.astimezone(utc).timestamp())
+    date_text = f"*Date and Time*\n<!date^{timestamp}^{{date_pretty}} at {{time}}|{penny_chat.date}>"
+
+    start_date = penny_chat.date.astimezone(utc).strftime("%Y%m%dT%H%M%SZ")
+    end_date = (penny_chat.date.astimezone(utc) + timedelta(hours=1)).strftime("%Y%m%dT%H%M%SZ")
+    google_cal_url = f'https://calendar.google.com/calendar/render?action=TEMPLATE&text=' \
+        f'{urllib.parse.quote(penny_chat.title)}&dates=' \
+        f'{start_date}/{end_date}&details=' \
+        f'{urllib.parse.quote(penny_chat.description)}'
```
JnBrymn: nitpick - split up the lines so that the url params are at the start of each line 
```
google_cal_url = (
    'https://calendar.google.com/calendar/render?'
    'action=TEMPLATE'
    f'text={urllib.parse.quote(penny_chat.title)}&'
    f'dates='{start_date}/{end_date}&'
    f'details={urllib.parse.quote(penny_chat.description)}'
)
```
(commas intentionally missing - it's a weird string concatenation syntax)

JnBrymn: the [calendar events api](https://developers.google.com/calendar/v3/reference/events) mentions a `hangoutLink` parameter. Maybe see if we can create default google hangout links? up to you

bot/processors/pennychat.py
```diff
@@ -0,0 +1,402 @@
...
+
+    return [
+        {
+            "type": "section",
+            "text": {
+                "type": "mrkdwn",
+                "text": f"*<{penny_chat.user}> invited you to a new Penny Chat!*"
```
JnBrymn: you need to make this print the user's name again

bot/processors/pennychat.py
nchouard: It looks like we would need to authenticate with google to do so, which I think is more work than it is worth right now.

bot/views.py
```diff
@@ -52,3 +54,31 @@ def interactive(request):
...
+
+    command = command_and_args[0]
+    arguments = ''
+    if len(command_and_args) > 1:
+        arguments = command_and_args[1]
+
+    if command == 'chat':
+        PennyChatBotModule.create_penny_chat(slack_client, event)
+    elif command == 'help':
+        print('no! I wont help you with ', arguments)
```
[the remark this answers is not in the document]
nchouard: I added a real help message instead :)

JnBrymn:
merge it! merge it! merge it!

PR #48, opened 2019-10-11
Move User model to new users app.

JnBrymn:
Doing this was confusing! How do you move a model to a new app and not risk migrations destroying data. I ended up going with basically [these instructions](https://fleschenberg.net/moving-a-django-model-to-another-app/) where we edit the migration files manually to make sure that the table gets moved rather than on app deleting the table and another one creating a new table of the same name.

Here's proof that I could do this in dev non-destructively:

```sh
# start on master branch
$ git checkout master

# remove database and create, run migrations (prior to the changes of this PR), and add some data
$ rm -fr db.sqlite3
$ ./manage.py migrate
Operations to perform:
  Apply all migrations: admin, auth, bot, contenttypes, sessions
...
$ ./manage.py shell -c "from bot.models import User; User.objects.create(real_name='bob'); print(User.objects.first().real_name)"
bob
# ^ user "bob" successfully created
```
At this point we just have a known database state on the master branch. Now to show that my changes work in the PR branch. In this branch I have already created a `users` app, I've moved the `User` model from `bot` to `users`, and I've created and manually edited the migrations.

```
$ git checkout extract-users-from-bot-4   # <- this was my 4th attempt!
$ ./manage.py migrate  # all the migrations are already made and in this branch
Operations to perform:
  Apply all migrations: admin, auth, bot, contenttypes, sessions, users
...

# now see if I can retrieve the preexisting data through the new users.models.User
$ ./manage.py shell -c "from users.models import User; print(User.objects.first().real_name)"
bob
$ ^ there it is!
```

After getting a 🚢 on this I'll do a production database backup and then push this branch to production and 🤞 hope it works. If it doesn't then I'll figure out how to upload the db backup.


users/models.py
```diff
@@ -0,0 +1,17 @@
...
+    real_name = models.CharField(max_length=100)
+    metro_name = models.CharField(max_length=200)
+    topics_to_learn = models.CharField(max_length=1500)
+    topics_to_share = models.CharField(max_length=1500)
+    how_you_learned_about_pennyu = models.CharField(max_length=500)
+    created = models.DateTimeField(auto_now_add=True)
+    updated = models.DateTimeField(auto_now=True)
+
+    class Meta:
+        db_table = 'user'
```
nchouard: This might be silly, but can we change this class and table name to "SlackUser"? It will allow us to differentiate users on slack from eventual Django users. Hopefully that will still play nicely with migrations.

JnBrymn: Let's talk about it. My immediate reaction is to not do this because _most_ of the data in this table corresponds to the deeper notion of a `User` - e.g. their email, topics to learn, topics to share, real name. In the future I could imagine having a `SlackUser` table or `Accounts` table or something that hold `slack_id, user_name, user_id` (the last of which being the foreign key to the user table. *Let's talk about it though. I don't completely understand your idea yet.*

I was thinking about changing the name of the table to `users_user` since that's what Django apparently wants me to do. I'd be interested on your thoughts here too.

nchouard: The reason I would like it to be at least renamed to something other than `User` is that django already provides a `User` model out of the box. https://docs.djangoproject.com/en/2.2/topics/auth/default/

nchouard: But yes, name of the table should follow Django's convention: `appname_modelname`

PR #51, opened 2019-10-20
Server static files in production

nchouard:
* Added Whitenoise to server static files
* Added gunicorn to serve Django app, as suggested by heroku docs https://devcenter.heroku.com/articles/django-app-configuration

Heroku automatically calls "collectstatic" which allows the files to be served in production.

### DIFF
--- a/Procfile
+++ b/Procfile
@@ -1,2 +1,2 @@
 release: python manage.py migrate
-web: gunicorn penny_university.wsgi
+web: gunicorn --workers=2 penny_university.wsgi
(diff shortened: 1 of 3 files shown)


nchouard:
I should have mentioned in my comments that I did test it in in a dev application. We are not running multiple workers yet. The default for Gunicorn is 1, which should be fine for now, but I just found out [Heroku actually recommends 2-3 workers for a hobby app](https://devcenter.heroku.com/articles/optimizing-dyno-usage#python). Should I go ahead and set it to two workers?

JnBrymn:
Sure, set a few workers on it if you'd like. The thing to remember though is that we can no longer cache a session in the django process. I don't think we're doing that yet. But, say, if we wanted to create a conversation bot that had to remember the history of the conversation, we would have to persist the conversation state in the db or in redis or something. No problem for now though.

PR #62, opened 2019-11-12
getting rid of unused endpoint, fix bug, nitpick

JnBrymn:
see comments

### DIFF
--- a/bot/processors/filters.py
+++ b/bot/processors/filters.py
@@ -41,7 +41,7 @@ def has_callback_id(callback_id, event):
     if 'callback_id' in event:
         return event['callback_id'] == callback_id
     else:
-        return event['view']['callback_id']
+        return event.get('view', {}).get('callback_id') == callback_id
 
 
 @event_processor_decorator
--- a/bot/processors/pennychat.py
+++ b/bot/processors/pennychat.py
@@ -172,10 +172,6 @@ def shared_message_template(penny_chat, user_name):
     ]
 
 
-def penny_chat_blocks(slack=None, penny_chat=None):
-    return save_message_template(slack, penny_chat)
-
-
 def penny_chat_details_modal(penny_chat):
     tz = timezone(penny_chat.user_tz)
     date = str(penny_chat.date.astimezone(tz).date())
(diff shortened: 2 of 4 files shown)


bot/processors/filters.py
```diff
@@ -41,7 +41,7 @@ def has_callback_id(callback_id, event):
     if 'callback_id' in event:
         return event['callback_id'] == callback_id
     else:
-        return event['view']['callback_id']
+        return event.get('view', {}).get('callback_id') == callback_id
```
JnBrymn: in the new bot framework, returning `True` means "pass the original event on", but if you pass a non-boolean, it's assumed that that's a new event to pass on in place of the original one - so the `else` clause effectively replaced the event with the string `event['view']['callback_id']`

bot/processors/pennychat.py
```diff
@@ -320,11 +316,15 @@ def create_penny_chat(cls, slack, event):
...
-                                                                        'date': date})
+        penny_chat, created = PennyChat.objects.get_or_create(
+            user=user['id'],
+            status=PennyChat.DRAFT_STATUS,
+            defaults={
+                'user_tz': user['tz'],
+                'template_channel': event['channel_id'],
+                'date': date,
+            },
+        )
```
JnBrymn: This is a nitpick formatting thing. Somehow it just works better for my eyes b/c it keeps everything shoved towards the left and it seems to better obey the python indenting thing. ¯\_(ツ)_/¯

bot/views.py
```diff
@@ -24,7 +24,7 @@ def index(request):
 @csrf_exempt
 def hook(request):
     blob = json.loads(request.body)
-    logging.info(blob)
+    logging.info(f'HOOK> {request.body.decode("utf-8")}')
```
JnBrymn: the logging is still pretty amateur, but here and below it's at least a little more descriptive.

bot/urls.py
```diff
@@ -6,6 +6,5 @@
     path('', views.index, name='index'),
     path('hook/', views.hook, name='hook'),
     path('interactive/', views.interactive, name='interactive'),
-    path('pennychat/', views.penny_chat, name='pennychat'),
```
JnBrymn: didn't need this no more

bot/views.py
```diff
@@ -41,25 +41,16 @@ def hook(request):
...
-    message_logger = logging.getLogger('messages')
-    message_logger.info(request.POST['payload'])
-
-    bot(payload)
-
-    return HttpResponse('')
-
-
-@csrf_exempt
-def penny_chat(request):
```
JnBrymn: didn't need this no more

Issue #68, opened 2019-11-14
Require a chat to be shared with SOMEONE


JnBrymn:
# TODO
TBD

# Description
Currently a `/penny chat` user can save and "share" a Penny Chat that isn't shared with any user or channel. Read up on ["Displaying Errors in Views"](https://api.slack.com/surfaces/modals/using#closing_views#displaying_errors) and fix it.

PR #69, opened 2019-11-15
error in penny chat is created w/ no invitees

JnBrymn:
*Goal:* if someone creates a penny chat w/ no channels or users invited, then error and provide a message to include some invitees

*Strategy:* In order to do this we have to respond to submission with a `response_action` payload. This means transmitting the message back through the `BotModule` and the `Bot` to the original call in `interactive` view handler. At both levels I'm choosing to send the event to all sub botmodules or event_handlers and if any _one_ of them return a non-`None` response then I'll propagate that back up the the original caller. If more than one returns a response then that's ambiguous (which response do you use?) so I raise and exception.

*Problem:* Slack's API is ... difficult. For `reasons` I don't seem to be able to attach error messages to the new action family of blocks. So I can't attach an error to the empty `invitees` and `channels` blocks in questions. There's also apparently no way to make a form-wide error (e.g. the message must be attached to one of the form elements). So just to get something out, I've stuck the error on the Description./e (The Title has scrolled off the top of the page.)

image

### DIFF
--- a/bot/processors/base.py
+++ b/bot/processors/base.py
@@ -150,7 +150,7 @@ class Bot:
         self.event_processors = event_processors
 
     def __call__(self, event):
-        """Sends event to all modules and returns response to caller.
+        """Sends event to all modules and optionally returns response to caller.
 
         Errors if there is more than on response from a module.
         :param event:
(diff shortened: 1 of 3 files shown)


JnBrymn:
closes https://github.com/penny-university/penny_university/issues/68

PR #74, opened 2019-11-24
penny university google group extraction script

JnBrymn:
* provides a date ordered list of chats where each chat is a date ordered list of the messages
* normalizes and fixes content
  * email addresses are modified to use the one that the person uses in slack
* things it can't do:
  * properly associate a URL with the text that it was originally linked to (information was missing) - so I just placed the URL inline with the text in parentheses
  * deal with images - I just ignored them or linked to them depending on what information was available
  * deal with code snippets - the code is still there but we don't know where the code block starts and ends

### DIFF
--- a/script/extract_google_forum_content.py
+++ b/script/extract_google_forum_content.py
@@ -1,0 +1,164 @@
+#!/usr/bin/env python3
+import argparse
+import ast
+from dateutil.parser import parse
+import json
+import mailbox
+import os
+import re
+
+codepoint_re = re.compile(r'=([0-9A-Fa-f]{2})')
+equal_then_space_re = re.compile(r'=\s+')
+weird_newline_re = re.compile(r'=\r?\n')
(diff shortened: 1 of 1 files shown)


script/extract_google_forum_content.py
```diff
@@ -0,0 +1,164 @@
...
+    args = parser.parse_args()
+
+    mbox_path = os.path.join(args.data_dump_path, 'topics.mbox')
+    print(mbox_path)
+    mbox = mailbox.mbox(mbox_path)
+    messages = get_messages(mbox)
+    messages = special_case(messages)
+    chats = get_chats(messages)
+    with open(args.output_file, 'w') as f:
+        f.write(json.dumps(chats))
```
nchouard: Picky: New line at end of file?

JnBrymn: This file ends in a new line. Typically the Python linter wants there to be a new line at the end of python files. Help me understand if I got something wrong here. Not seeing it.

PR #79, opened 2019-11-29
React Frontend

### LINKED ISSUE #77: Create React Application
Create a simple react application for the forum. The application will use some dummy data, and all the components should use props rather than state (in order to plug in Redux later).

nchouard:
Closes #77 -- create a React frontend app. It is not hooked up to our Django app at the moment, and only uses mock data. Also added docs for how to run the frontend as it currently stands. There may be some files we don't need in here, like `serviceWorker.js`, but I haven't dug into whether we want to use it or not so I am leaving it for now.


nchouard:
@JnBrymn I'm gonna go ahead and merge it but even if you we were to push the code up to heroku, we wouldn't have a way of viewing it because there no build process built in to connect it to our django app and serve it yet.

Issue #82, opened 2019-12-04
Create Architecture for API


JnBrymn:
closed by https://github.com/penny-university/penny_university/pull/84

- [ ] Update "participants" for chats

PR #84, opened 2019-12-04
Api design

JnBrymn:
closes https://github.com/penny-university/penny_university/issues/82

# TODO:
- [ ] Figure out how to represent participants of the penny-chats instead of a single "user".


docs/achitecture_and_design/api_design.md
```diff
@@ -0,0 +1,121 @@
...
+            "follow_up_count": 5
+        },
+        ...
+    ]
+}
+```
+
+### Retrive all chats
+`GET /api/chats`
+(Same data returned as User's Chats)
```
wgaggioli: these are the chats for the currently logged in user?

docs/achitecture_and_design/api_design.md
```diff
@@ -0,0 +1,121 @@
...
+    "previous": "...url to next previous of results...",
+    "follow_ups": [
+        {
+            "user": {
+                "id": 1,
+                "name": "..."
+            },
+            "date": "...",
+            "content": "...",
+            "penny_chat": "api/chats/<id>"
```
Raab70: Not just the `<id>`?

docs/achitecture_and_design/api_design.md
```diff
@@ -0,0 +1,121 @@
...
+
+
+### Delete a follow-up
+`DELETE /api/followups/<id>`
+
+
+## For later discussion:
+
+### Chat search
+`GET /api/chats/search?user=2`
```
Raab70: I've used filtering but not search plugins and usually it's not a dedicated `/search` endpoint but just parameters on the listing endpoint. Like so: https://www.django-rest-framework.org/api-guide/filtering/#django-rest-framework-full-word-search-filter

docs/achitecture_and_design/api_design.md
```diff
@@ -0,0 +1,121 @@
...
+`GET /api/user/2/chats`
+```json
+{
+    "count": 123,
+    "next": "...url to next page of results...",
+    "previous": "...url to next previous of results...",
+    "chats": [
+        {
+            "id": 1,
+            "user": {
```
sburns: A single user owns the chat? Is there a M2M for "involved people" and would you expect to encode that here or in the ChatDetail?

JnBrymn: This comment ended up generating a lot of discussion in our hack tonight regarding both the API and the underlying data model. We're thinking about moving away from `user` and to `participants`.  Where a participant is your M2M relationship between user and chat, but we will probably also have a type associated with the participant which can be CREATOR, INVITEE, ATTENDEE, and maybe COMMENTER. (Can you think of others?)

docs/achitecture_and_design/api_design.md
JnBrymn: In our current product, all chats are public and `/api/chats` will simply return all chats that exist. Obviously this will be a big number at some point and we'll do away with this. But as stated here, it represents a replacement for [our forum](https://groups.google.com/forum/#!forum/penny-university) which is a list of all chats.

docs/achitecture_and_design/api_design.md
JnBrymn: It's not obvious with the way we've written it, but all the relationships will be represented with hyperlinks (like in the github v3; ex: api https://developer.github.com/v3/issues/#response ).

docs/achitecture_and_design/api_design.md
JnBrymn: We'll get rid of `/search` then.

docs/achitecture_and_design/api_design.md
JnBrymn: COMMENTER might not be a necessary type since the follow-ups (comments) will have users associated with them.

PR #89, opened 2019-12-24
Rest API

### LINKED ISSUE #80: Basic REST API
Create a basic REST API to get data to our frontend. No user auth is necessary yet!

https://hackmd.io/VycYERAeQFW0pWVLluszhg?both

- [x] GET/POST Penny Chats (`/chats`)
- [x] GET/PUT/DELETE Penny Chat (`/chats/{id}`)
- [x] GET/POST Follow Ups (`/chats/{id}/followups`)
- [x] PUT/DELETE Follow Ups (`/followups/{id}`)
- [x] Rebase master to get real data
- [x] Add and connect user profile serializer
- [x] Tests

nchouard:
closes #80


pennychat/tests/utils.py
```diff
@@ -0,0 +1,42 @@
...
+
+from django.utils import timezone
+
+from pennychat.models import PennyChat, FollowUp
+from users.models import UserProfile
+
+
+def generate_user():
+    user, created = UserProfile.objects.get_or_create(real_name='Anonymous Profile',
+                                                      defaults={'email': 'anonymous@profile.com'})
```
JnBrymn: nit pick - could I talk you into standardizing around using this code format
```python
user, created = UserProfile.objects.get_or_create(
    real_name='Anonymous Profile',
    defaults={'email': 'anonymous@profile.com'},
)
```
rather than the above? The rationale is that the arguments are always one more indent than the command that created the arguments. If you back up and look at the indenting in this entire file the arguments are indented much differently for every function here. Another benefit of my recommended format is that it avoids arguments getting piled up at the right side of the screen as seen above.

pennychat/views.py
```diff
@@ -1,3 +1,77 @@
...
+
+    def perform_create(self, serializer):
+        # get or create an anonymous user since we don't have auth
+        # this will eventually be replaced by user = self.request.user
+        user, created = UserProfile.objects.get_or_create(
+            real_name='Anonymous Profile',
+            defaults={
+                'email': 'anonymous@profile.com'
+            })
+        serializer.save(user=user)
```
JnBrymn: This confuses me. It seems to save a penny chat with no content and it attaches it to an anonymous user. I don't see how that will ever be useful. I'm not going to block shipping on this, it just represents something that I don't understand yet.

pennychat/views.py
```diff
@@ -1,3 +1,77 @@
...
+    def get(self, request, pk, format=None):
+        follow_ups = FollowUp.objects.filter(penny_chat_id=pk)
+        queryset = self.filter_queryset(follow_ups)
+
+        page = self.paginate_queryset(queryset)
+        if page is not None:
+            serializer = self.get_serializer(page, many=True)
+            return self.get_paginated_response(serializer.data)
+
+        serializer = self.get_serializer(queryset, many=True)
```
JnBrymn: nit pick: `self.get_serializer(queryset, many=True)` this seems like it's doing something, but in reality it's just returning an empty `queryset`, right? Alternatives: Does this code work `self.get_serializer(None, many=True)`? If so you don't need the if clause. How about an empty list? `self.get_serializer([], many=True)` In that case you can replace line 39 with `page = self.paginate_queryset(queryset) or []` and remove the if clause. Finally if that doesn't work you can just do this:

```python
        page = self.paginate_queryset(queryset) or queryset
        serializer = self.get_serializer(page, many=True)
        return self.get_paginated_response(serializer.data)
```
just one less `if` clause to worry about

pennychat/tests/utils.py
nchouard: Yes, definitely!

penny_university/settings.py
```diff
@@ -40,10 +40,12 @@
...
-    'bot.apps.BotConfig',
-    'home.apps.HomeConfig',
-    'pennychat.apps.PennychatConfig',
-    'users.apps.UsersConfig',
+    'bot',
+    'rest_framework',
+    'api',
+    'home',
+    'pennychat',
+    'users',
```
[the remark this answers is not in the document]
nchouard: I had never seen that before, but makes sense to me. I'll switch it back!

pennychat/tests/api/test_penny_chat.py
```diff
@@ -0,0 +1,93 @@
...
+
+
+@pytest.mark.django_db
+def test_follow_up_url(chats_setup, client):
+    response = client.get('/api/chats/1/')
+    follow_up_url = response.data['follow_ups']
+    assert follow_up_url == 'http://testserver/api/chats/1/follow-ups/'
+    response = client.get(follow_up_url)
+    assert response.status_code == 200
+    assert response.data['count'] == 0
```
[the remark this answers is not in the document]
nchouard: Yeah, at first I overlooked this because I was creating tests for the model view set, in which it would be an obvious outcome to save the model. However, there are places I am saving manually, so I think it makes sense to add this on all of them.

pennychat/tests/utils.py
sburns: Y'all might consider using the black code formatted and verifying that style in CI.

JnBrymn: Black is worth considering. I like what gofmt has done to golang development. Black would be the same thing to python. Last time I looked at Black, I didn't like some of the defaults and the defaults couldn't be changed - namely the preference for double quotes. Perhaps I should reconsider.

pennychat/views.py
nchouard: `perform_create` is called in the background from a post in a ModelViewSet. The serializer object here will have all the relevant data that is needed for the chat. I attach the anonymous user for now since there is no other way of figuring out who is created the chat without auth.

pennychat/views.py
nchouard: You were right, no need for the if clause!

Issue #102, opened 2020-01-09
Fix Dev passthrough dying


JnBrymn:
The dev passthrough dies if a request takes too long to return. It generates an error like `2020-01-09T04:40:13.556749+00:00 app[web.1]: [2020-01-09 04:40:13 +0000] [4] [CRITICAL] WORKER TIMEOUT (pid:21)` and then restarts the server, thus removing the passthrough. This is annoying. Fix it.

PR #104, opened 2020-01-13
timing out passthrough at 25 seconds

JnBrymn:
closes https://github.com/penny-university/penny_university/issues/102


JnBrymn:
Low probability of problems because this only deals with the development passthrough that I use.

PR #105, opened 2020-01-13
Updated import commands

### LINKED ISSUE #99: Make it possible to pipe file into django import_google_forum
(Closed by #105)

Make it possible to pipe file into django import_google_forum so that we can use it from heroku w/o shipping data in.

information here: https://stackoverflow.com/a/49152992/348056

JnBrymn:
closes #99 

## import_google_forum
* Must now be run by piping in the .mbox rather than supplying the path of the file. This will allow us to run the command directly against Heroku like `cat /wherever/topics.mbox | heroku run --no-tty -a <your-app> ./manage.py import_google_forum --to_database`
* Removed the user input confirmation and replaced it with a new `--live_run` arg. In both cases the output of the method lists all new objects that will be created, but this will only write to the database if `--live_run` is supplied.
* Weird note, `--live_run` can't be the last argument or command will fail for some reason. It seems to be a Heroku problem because the error is coming from a .js file on their side.

## import_users_from_slack
* Removed the user input confirmation and replaced it with a new `--live_run` arg so that it was consistent.

### DIFF
--- a/README.md
+++ b/README.md
@@ -65,6 +65,7 @@ To cancel just visit http://yourapp.herokuapp.com/forward
 
 
 ## Pull in data from forum or slack
-* Pull chats and follow ups from Google Group by getting a dump of the data from https://takeout.google.com, unzipping it, and then running `./manage.py import_google_forum --data_dump_path <somewhere>/Takeout/Groups/penny-university\@googlegroups.com/topics.mbox  --to_database`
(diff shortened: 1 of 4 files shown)


JnBrymn:
# Evidence of correctness
I deployed to master and ran both the user and pennychat import commands as currently documented in the README and they worked as expected. Merging!

PR #106, opened 2020-01-14
Add state design docs

nchouard:
I wanted to create some documentation to describe the state of the frontend application, [like they do here in the redux docs](https://redux.js.org/basics/reducers#designing-the-state-shape). This is just a start taking into consideration how the API is built, and I'm sure it will change over time. Please provide any feedback you may have!


JnBrymn:
Nice! I like the idea of doc-driven development for these high-level interface-related questions.

Thing I'll ask about tomorrow night: do we really need the `byId` layer? Or, say, could `chats` directly be a object with `id` for keys.

PR #108, opened 2020-01-20
fixed pennychat import status bug

JnBrymn:
# Todo
- [ ] Send it 🚢 
- [ ] delete all penny chats in the production db and re-run the import (this guarantees that all those penny chats have status=COMPLETED)

# Description
I was trying to track down a bug in production where /penny_chat errors by printing to a non-existent channel. Found a related bug where all of the imported penny chats had the default status = draft. That's bad because we're grabbing currently grabbing the draft penny chat and using it for our `/penny chat` modal.

This change creates a new `completed` status and assigns all imported pennychats with that status.

As soon as I send this I think I'll blow away the production PennyChats and rerun the import. We're not worried about losing Penny Chats yet are we?

See update: https://github.com/penny-university/penny_university/pull/108#issuecomment-576516128


nchouard:
Sounds good, but looks like some tests are failing. Is there a setting that needs to be changed somewhere?

JnBrymn:
# Update
I figured out the original bug. The reason that we were getting channel not found errors is because the the `/penny chat` command was initiated in a direct message channel between two users and Penny doesn't have access to these channels. I just made it fall back to sending the message to the originating user's slackbot channel.

PR #115, opened 2020-02-05
Make deployment docs

JnBrymn:
We need to start keeping track of docs for things like this. This is a start. I think some items here aren't sound, so let's revise this first and then ship it.


docs/deployment.md
```diff
@@ -0,0 +1,16 @@
+# Deployment From PR to Production
+
+1. Create a PR. Get it reviewed.
+2. **Make sure master is merged into the PR branch** and make sure all CI tests pass.
+3. Deploy to your QA slack and app environment first. (Something like `git push qa my_pr_branch:master`)
```
JnBrymn: I feel like something more is needed before 3 in order to make sure that QA is actually first in the same state as PROD master - especially as it involves migrations.

docs/deployment.md
```diff
@@ -0,0 +1,16 @@
...
+4. Watch the logs and manually tests your changes in QA.
+5. Once you're satisfied that it's safe deploy the PR branch to PROD. (Something like `git push production my_pr_branch:master`)
+6. Watch the logs and manually tests your changes in PROD.
+7. If all goes well, then merge your branch into master and you're done!
+
+## Rollback
+If something horrible happens the `git push -f production master`
```
JnBrymn: This seems sloppy and I also worry about migrations issues.

docs/deployment.md
```diff
@@ -0,0 +1,16 @@
+# Deployment From PR to Production
+
+1. Create a PR. Get it reviewed.
+2. **Make sure master is merged into the PR branch** and make sure all CI tests pass.
```
Raab70: I get that you're documenting a process but on my team we prefer to rebase PRs on master to avoid merge commits all over the place. It forces the PR author to clean up the code for their PR to match changes necessary for master.

docs/deployment.md
Raab70: Yeah I'd love to see more exact commands. For example step 1: `export MY_BRANCH=<your-branch-name>` Then your commands can be copy-pasted exactly, `git push qa $MY_BRANCH:master`

docs/deployment.md
Raab70: Not with migrations that I've ever seen... So let's give it a go. You could maybe do some tricky scripting, something like: `python manage.py migrate my_app $(ls -t my_app/migrations/ | head -1 | cut -d . -f 1)` This should target the most recent migration, not sure if this would work in your deployment, does GIT copy file modification dates on clone? You could always just sort by filename instead. But beware then you won't pull in migrations from other apps, like if you add a package that has an app it may need migrations. Also, if your migrations are non-elidable you're just screwed of course.

docs/deployment.md
JnBrymn: I used to be pretty hardline on this myself, but I was amazed that no one at GitHub seemed to care ¯\_(ツ)_/¯ - I'm trying the koolaid myself right now... kinda nice.

docs/deployment.md
JnBrymn: added

docs/deployment.md
JnBrymn: For now I added docs about how you could by hand. To automate this, I think I would do some sort of filename-only version of `git diff master` and look for all new migration files.

That said... it might just be better to always do this by hand. Migrations always feel special case enough that automation might just hurt more.

PR #119, opened 2020-02-10
created invitation model

### LINKED ISSUE #118: Split PennyChat into Chat part and new Invitation model
# Context
The current PennyChat model is a combination of stuff that is important to the chat, and then extra fields that are only used at creation time and _then_ only because the Slack API is lacking and we're compensating.

# Goal
Split model into PennyChat and Invitation. The invitation will handle all the annoying fields described above plus anything else we need to keep state during the invitation phase. For instance, we'll need to keep track of message ids where the invitation has been posted so that we can update them, but once the PennyChat is complete, we won't need this information ever again.

JnBrymn:
This introduces an invitation model that will keep all the Slack API and invitation-time-only kruft out of the longterm PennyChat model. If we don't do this, then I fear using some of the Invitation only kruft in the application which will make it hard to remove later.

closes #118 
Note that I took a different approach than I said I would take in #118. Instead of only putting the SlackAPI and invite-time-only junk into the Invitation, I also included the PennyChat title, description, and date. This data is now redundant once the real PennyChat gets created. But if I had gone with the #118 approach, we would have to constantly refer to both the Invitation and the PennyChat to get the information that we need during creation. This approach was easier, and doesn't block us from fully going with #118 later.

# TODO
- [ ] Review
- [ ] Deploy
- [ ] Manually update the statuses (`status = status * 10`)


bot/tests/processors/test_pennychat.py
```diff
@@ -77,6 +84,69 @@ def test_time_select(mocker):
...
+    def id_mock(user_id, slack_client):
+        return ids_mock([user_id], slack_client).get(user_id)
+
+    with mocker.patch('bot.processors.pennychat.get_or_create_user_profile_from_slack_ids', side_effect=ids_mock), \
+            mocker.patch('bot.processors.pennychat.get_or_create_user_profile_from_slack_id', side_effect=id_mock), \
+            mocker.patch('bot.processors.pennychat.requests'):
+
+        # The Actual Test
+        PennyChatBotModule(mocker.Mock()).share(event)
+
```
JnBrymn: if you share the event, then that's when we create the actually penny chat and participants

pennychat/models.py
```diff
@@ -6,19 +6,40 @@
 
 
 class PennyChat(models.Model):
-    DRAFT_STATUS = 1
-    SHARED_STATUS = 2
-    COMPLETED_STATUS = 3
+    SHARED = 20
+    COMPLETED = 30
+    ABANDONED = 40
```
JnBrymn: I multiplied all statuses by 10 so that we can fit things into the ranges if need be - for example, what if in the future there is more than one way to share a chat? It would be nice for that share to be `21` so that it's grouped with the existing share.

JnBrymn: also penny chats are created when the invitation is shared, thus penny chats don't need `DRAFT`

bot/processors/pennychat.py
```diff
@@ -31,32 +35,22 @@ def get_time_options():
     return time_options
 
 
-def shared_message_preview_template(slack_client, penny_chat):
+def shared_message_preview_template(slack_client, penny_chat_invitation):
```
JnBrymn: I removed Participant creation from here (it was just the preview)

bot/processors/pennychat.py
```diff
@@ -435,26 +436,63 @@ def share(self, event):
                 ),
             )
             return
-        user = self.slack.users_info(user=penny_chat.user)['user']
-        for share_to in penny_chat.channels.split(',') + penny_chat.invitees.split(','):
+
```
JnBrymn: I added the _real_ PennyChat creation and  Participant creation from here (it is the share - that seems a better place to make this real)

JnBrymn:
Updated postgres db with `update pennychat_pennychat set status = status * 10 where status < 10;
UPDATE 100`

```

penny-university::DATABASE=> select status from pennychat_pennychat where status < 10;
 status 
--------
      2
      2
      2
      2
      2
      3
      3
      3
      3
      3
      3
      3
      3
      3
      3
      3
      3
      3
      3
      3
      3
      3
      3
      3
      3
      3
      3
      3
      3
      3
      3
      3
      3
      3
      3
      3
      3
      3
      3
      3
      3
      3
      3
      3
      3
      3
      3
      3
      3
      3
      3
      3
      3
      3
      3
      3
      3
      3
      3
      3
      3
      3
      3
      3
      3
      3
      3
      3
      3
      3
      3
      3
      3
      3
      3
      3
      3
      3
      3
      3
      3
      3
      3
      3
      3
      3
      3
      3
      3
      3
      3
      3
      3
      3
      3
      3
      3
      3
      3
      3
(100 rows)

penny-university::DATABASE=> update pennychat_pennychat set status = status * 10 where status < 10;
UPDATE 100
penny-university::DATABASE=> select status from pennychat_pennychat where status < 10;
 status 
--------
(0 rows)

penny-university::DATABASE=> select status from pennychat_pennychat;
 status 
--------
     20
     20
     20
     20
     20
     20
     30
     30
     30
     30
     30
     30
     30
     30
     30
     30
     30
     30
     30
     30
     30
     30
     30
     30
     30
     30
     30
     30
     30
     30
     30
     30
     30
     30
     30
     30
     30
     30
     30
     30
     30
     30
     30
     30
     30
     30
     30
     30
     30
     30
     30
     30
     30
     30
     30
     30
     30
     30
     30
     30
     30
     30
     30
     30
     30
     30
     30
     30
     30
     30
     30
     30
     30
     30
     30
     30
     30
     30
     30
     30
     30
     30
     30
     30
     30
     30
     30
     30
     30
     30
     30
     30
     30
     30
     30
     30
     30
     30
     30
     30
     30
(101 rows)

```

PR #134, opened 2020-02-20
replace janky PennyChatInvitation

JnBrymn:
# WHAT
Replace my previous PennyChatInvitation with something more Django-esque

Still to do:
- [ ] Snapshot database and name is master's hash (so we know when it was last valid).
- [ ] Delete all draft PennyChats in production
- [ ] Delete ALL PennyChatInvitations because they will erroneously refer back to PennyChat #1 (due to having to set a default foreign key during migration)

# WHY
My previous PennyChatInvitation contained duplicate fields of title, description, date, and status. Whenever a PennyChatInvitation was shared, the "real" PennyChat was created and copied over the shared fields. *That caused a problem.* If we ever want to update the PennyChatInvitation fields, then we have to remember to also update the Penny Chat fields to make them consistent. Clearly we only needed a single canonical field for each of these redundant fields.

I found out that Django's [Multi-table-inheritance](https://docs.djangoproject.com/en/dev/topics/db/models/#multi-table-inheritance) is the perfect thing for this! All the code works about like before, but I could drop the crufty field copy-over I was doing.

This also unblocks me for doing PennyChatInvite editing.


sburns:
What does the `user` mean on the PennyChatInvitation? Is that who sends it or gets it? Or created the PennyChat? I'm probably missing something.

JnBrymn:
@sburns, the `PennyChatInvitation.user` used to be the organizer or the `PennyChat` and it's the slack_id, not the database_id. But now I think we're using it only in the `PennyChatInvitation` to have an easy way to contact the organizer in slack without doing `UserProfile.objects.get(participation__role='Organizer', participation__penny_chat=this_penny_chat).slack_id`... though 🤔 maybe that's the way to go?

JnBrymn:
I'm running into these type issues when attempting to push to QA/Prod https://code.djangoproject.com/ticket/26703

JnBrymn:
whew 😌  - solved it

For anyone interested, the problem was that I converted an existing model (`ThingDetails`) to a [Multi-table inheritance](https://docs.djangoproject.com/en/dev/topics/db/models/#multi-table-inheritance) model that subclassed a different existing model `Thing`. This implicitly creates a `models.OneToOneField` in the subclass that points to the parent with a hidden `parent_id` field that must be both non-null and unique. So the question is - what do you do with all the previously existing `ThingDetails` that now need a `parent_id` field? It can't be null, and it must be unique. Since there was no sensible default I couldn't make the migration work.

Fortunately I realized that I didn't care if the existing `ThingDetails` model burned in a glorious fire of a thousand suns. So I hand wrote a migration to drop the old `ThingDetails` model and then created a migration to introduce the new one. That resolved the problem. Since this told the migration engine that no prior fields had to be populated, everything worked fine.

Issue #136, opened 2020-02-23
Allow organizers to edit their PennyChats


JnBrymn:
# TODO
- [x] Upon sharing a Penny Chat
  - [x] The necessary data will be recorded so that the individual shares can be updated.
  - [x] The organizer will get a button to edit their event.
- [x] Upon clicking the edit button
  - [x] If text or title is edited then reshare the new information on all places it's already shared
  - [x] If more users are created or removed then you have to create new shares or delete old ones.
- [x] test
- [x] review #138
- [x] ship #138

JnBrymn:
Note to self 
```
# TODO! START in cheate_penny_chat - if there is for some reaons more than one invitation drafts then the user can no longer use /penny chat
        # because this errors. Fix by try/except around this and if there are problems, set all the STATUSES to WEIRD and create a new one
        # THEN you're ready for trying to get the penny chat by the id stored in value
```

JnBrymn:
I want to make a processor like this to stick in front of all the processors for `PennyChatBotModule`
```python
@event_processor_decorator
def gather_penny_chat_data(event):
    try:
        value = json.loads(event['actions'][0]['value'])
        penny_chat_invitation = PennyChatInvitation.objects.get(id=value['penny_chat_id'])
        event['penny_chat_data'] = value
        event['penny_chat_data']['penny_chat_invitation'] = penny_chat_invitation
        return event
    except:  # noqa
        import ipdb;ipdb.set_trace()
        requests.post(event['response_url'], json={'delete_original': True})
        self.slack_client.chat_postEphemeral(
            channel=event['channel']['id'],
            user=event['user']['id'],
            text=(
                "We are sorry, but an error has occurred and the Penny "
                "Chat you are trying to edit is no longer available."
            ),
        )
        return
```

JnBrymn:
all boxes checked!

PR #138, opened 2020-02-27
Make penny chat invitations editable

JnBrymn:
# What
Make penny chats editable. See my evidence of correctness https://github.com/penny-university/penny_university/pull/138#issuecomment-591792134

# Why
Because it's an obvious need.

# Related
See #136 for my sprawling incoherent notes.﻿


JnBrymn:
Create event and share
image

I see that it's shared in the channel it should be in
image

I also get this direct message saying that I shared and letting me have a button to edit it.
image

I click the edit button and change ALL the details including room and invitee (to be me since it's easy to check)
image

Now I get a direct share
image

Also my edit template pops back up with the correct information.

The invitation that was in general is no longer there.

There is now an invitation in the new channel with all the correct info
image

Clicking my edit button once more, and I see that the edit template disappears as it should.

If I close the resulting modal, the edit template pops back up again with the unaltered data, and nothing happens to the invitations posted in the channel.

PR #148, opened 2020-03-07
Create and Update New Follow Ups

nchouard:
We can now update and add new follow ups! Note that there is still no user model, so anyone can edit follow up. Also, I added some code that you can uncomment in order to be able to create new follow ups using an anonymous user. Follow ups can't be created if you don't uncomment those lines.

I also fixed the tests because we are no longer paginating follow ups, and added an ordering to participants to fix a warning in one of the tests.


penny_university_frontend/src/components/followups/FollowUpCard.js
```diff
@@ -1,12 +1,31 @@
...
+  }
 
-const FollowUpCard = ({followUp}) => {
   return (
     <div className='pt-2'>
-      <FollowUpUserInfo user={followUp.userInfo} date={followUp.date}/>
-      <Content className='ml-4 border-left pl-3' content={followUp.content}/>
+      <div className='d-flex justify-content-between'>
+        <FollowUpUserInfo user={followUp.userInfo} date={followUp.date}/>
+        {editMode ?
```
JnBrymn: Somehow this is eventually where we hook in and get whether or not this is the _authed_ user (e.g. the only one allowed to edit). Should we start creating tickets for things that will change after we have an authed user?

penny_university_frontend/src/reducers/paginate.js
```diff
@@ -6,6 +6,7 @@ const union = (a, b) => (
 // Reducer for handling actions that require pagination. Requires three types (request, success, and failure),
 // and a function to map the pagination to a key, e.g. (action) => action.filterName.
 const paginate = ({ types, mapActionToKey }) => {
+  console.log("HIT IT!")
```
JnBrymn: unhit it please :)

penny_university_frontend/src/components/chats/ChatDetail.js
```diff
@@ -2,10 +2,10 @@ import React, {useEffect} from 'react'
 import {HeartButton, CreateButton} from '../buttons'
 import {Card} from 'reactstrap'
 import Date from '../Date'
-import Content from '../Content'
+import Content from '../content/Content'
```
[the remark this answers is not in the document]
nchouard: Good catch! You can do either, but I prefer `import {Content} from '../content'`, but that's only possible when it's exported from a package. Originally Content.js was an orphan file like Date.js and I forgot to change it.

penny_university_frontend/src/reducers/paginate.js
nchouard: 🤣

penny_university_frontend/src/components/followups/FollowUpCard.js
nchouard: At this point, it's really only editing and creating follow ups. I'm not sure if we need separate tickets for each?

penny_university_frontend/src/components/chats/ChatDetail.js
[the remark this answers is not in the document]
nchouard: Just realized I did change it after all. It looks like you were reviewing based on only the first commit.

Issue #153, opened 2020-03-19
Set up `process_tasks` and move PennyChat slack messages into background tasks.


JnBrymn:
# TODO
- [x] Set up process_tasks and make sure it works locally.
- [x] Move the slack messaging functionality into bot.tasks.pennychat
- [x] More tests
- [x] Cleanup
  - [x] Remove TODO!
- [x] Figure out how to run locally
- [x] Update docs
- [x] Get working on Heroku (make sure the task runner works with 10 minute Heroku Scheduler)

# Description
Use [process_tasks](https://django-background-tasks.readthedocs.io/en/latest/)  so that we can call functions and automatically have them backgrounded. Use this in conjunction with the [Heroku Scheduler](https://devcenter.heroku.com/articles/scheduler) to restart the process_task daemon every 10 minutes.

JnBrymn:
# Proof of correctness
![image](https://user-images.githubusercontent.com/566533/78460552-5f509300-7687-11ea-8427-02beb27a384b.png)

I created the first Penny Chat before turning on the scheduled tasks and (as expected) no notifications were sent _until_ the scheduled task kicked in. After finally receiving the notification I created another penny chat and was notified almost immediately.

PR #155, opened 2020-03-29
Moved long-running invitation slack message postings to background task runner

JnBrymn:
# What
Moved long-running invitation slack message postings to background task runner

# Why
Because we have a 3 second deadline to return to the slack API, and these messages were taking an arbitrarily long time to send depending upon how many users were invited to a penny chat.

# Related
parent https://github.com/penny-university/penny_university/issues/153


bot/processors/pennychat.py
```diff
@@ -53,158 +55,6 @@ def get_time_options():
     return time_options
```
JnBrymn: Below moved to task/pennychat.py

PR #158, opened 2020-04-01
Backend User Auth & Refactor User -> Profile

### LINKED ISSUE #154: Refactor "user" to "profile

nchouard:
closes #154 

The rest API now enforces authentication for creates, updates and deletes. This update also includes the refactor of user to profile to reduce confusion between the django `User` and our `UserProfile`.


penny_university_frontend/src/reducers/index.js
```diff
@@ -15,6 +15,8 @@ const entities = (state = { chats: {}, followUps: {}, users: {} }, action) => {
 const error = (state = null, action) => {
   const { type, error } = action
 
+  console.log(error)
+
```
nchouard, line 19: bad

JnBrymn:
maybe ping me on slack once you have the "evidence of completeness" in place. I'm bad about checking my github notifications

pennychat/views.py
```diff
@@ -59,12 +74,19 @@ class UpdateDeleteFollowUp(mixins.UpdateModelMixin, mixins.DestroyModelMixin, ge
     """
     queryset = FollowUp.objects.all()
     serializer_class = FollowUpSerializer
+    permission_classes = [IsAuthenticated]
 
     def put(self, request, *args, **kwargs):
+        if self.get_object().user_profile.user != request.user:
```
CoreyAR, line 80: This could also get put into a mixin for this class  if every action is protected against unauthenticated users

nchouard:
# Evidence of completeness

## UserProfile is now distinct from User in the API
Participants and Follow Ups now reference a `user_profile` which points to a user. Here's an example response for a chat detail:
```json
{
    "id": 4,
    "url": "http://localhost:8000/api/chats/4/",
    "title": "Math with jnbrymn",
    "description": "",
    "date": "2017-01-17T18:28:52Z",
    "follow_ups": "http://localhost:8000/api/chats/4/follow-ups/",
    "participants": [
        {
            "user_profile": {
                "id": 5,
                "url": "http://localhost:8000/api/users/5/",
                "real_name": "Jason Orendorff",
                "user": null,
                "chats": "http://localhost:8000/api/users/5/chats/"
            },
            "role": "Organizer"
        },
        {
            "user_profile": {
                "id": 4,
                "url": "http://localhost:8000/api/users/4/",
                "real_name": "John Berryman",
                "user": null,
                "chats": "http://localhost:8000/api/users/4/chats/"
            },
            "role": "Attendee"
        }
    ]
}
```

## Register
You can now register a new user. Any profiles associated with the provided email will be attached to that user.
register
user_profile

## Log In
There is a login endpoint to retrieve an authentication token.
login

## Authenticated Requests
You can use the `key` returned from the login or register requests to make authenticated requests to the API. Creating, updating, and deleting chats and follow ups require authentication. Add the `Authorization` header with the token to requests in order to authenticate the user.
register

#### Permissions
If a user is attempting to update a follow up or penny chat to which they do not have access, it will return a Forbidden response.
forbidden

## Log Out
There is also a log out endpoint, which clears the token that you pass in as an authentication header. The user will no longer be authenticated and will have to log in again to receive a new token.
logout

pennychat/views.py
```diff
@@ -16,8 +19,26 @@ class PennyChatViewSet(viewsets.ModelViewSet):
     serializer_class = PennyChatSerializer
 
     def perform_create(self, serializer):
-        # TODO add participants and make request user's UserProfile as the Organizer
-        serializer.save()
+        if not self.request.auth:
```
CoreyAR, line 22: It looks like this whole class is authenticated so we could use a permissions class.

pennychat/views.py
```diff
@@ -34,20 +55,15 @@ def get(self, request, pk, format=None):
...
-            # creating new follow ups will not work unless you uncomment these lines!
-            # user, created = UserProfile.objects.get_or_create(
-            #     real_name='Anonymous Profile',
-            #     defaults={
-            #         'email': 'anonymous@profile.com',
-            #         'slack_team_id': '1'
-            #     })
-            # serializer.save(user=user)
+            profile = UserProfile.objects.get(user=request.user)
+            serializer.save(user_profile=profile)
```
JnBrymn, line 66: This will be weird at some point. When a user has more than one slack then they will have more than one UserProfile and this will actually cause an error and they will not be able to post FollowUps. Definitely don't have to solve this now, but I am puzzling about how this will change going forward. 

https://github.com/penny-university/penny_university/issues/168

docs/achitecture_and_design/api_design.md
```diff
@@ -110,7 +121,9 @@ The following are the RESTful endpoints needed to support basic chat and follow-
         {
             "user": {
                 "id": 1,
-                "name": "..."
+                "real_name": "...",
+                "user": {...},
+                "chats": "http://pennyuniversity.org/api/users/1/chats/"
             },
             "date": "...",
             "content": "...",
```
JnBrymn, line 129: thanks for updating docs!

pennychat/views.py
nchouard: The list and detail are not for this one

pennychat/management/commands/import_google_forum.py
```diff
@@ -215,39 +215,39 @@ def get_chats(messages):
...
         followups = []
         for message in chat['messages']:
             followups.append({
                 'content': message['body'],
                 'date': message['date'],
-                'user': message['from'],
+                'user_profile': message['from'],
             })
-        formated_chats.append({
+        formatted_chats.append({
```
JnBrymn, line 227: lol - thanks

pennychat/views.py
CoreyAR: I actually meant to say that we could inherit from a base viewset that overrides get_object/ get_queryset to prevent users from accessing objects  that don't have access to.

pennychat/views.py
CoreyAR: It might make sense to use a decorator then to encapsulate the logic

pennychat/views.py
nchouard: Ah! I like that idea.

pennychat/views.py
nchouard: That makes sense, I'll do that.

pennychat/permissions.py
```diff
@@ -0,0 +1,10 @@
+from rest_framework.permissions import BasePermission
+
+
+class IsOwner(BasePermission):
+    """
+    Allows access only to the user who owns the entity.
+    """
+
+    def has_permission(self, request, view):
+        return bool(view.get_object().user_profile.user == request.user)
```
JnBrymn, line 10: I'm interested in understanding how this works. The logic looks right of course. But what is `get_object()` doing? This will bread on some views certainly, right? (those w/o a user_profile) Does this need a little more documentation?

pennychat/tests/api/test_follow_up.py
```diff
@@ -36,9 +41,24 @@ def test_create_follow_up(test_chats_1):
 
 
 @pytest.mark.django_db
-def test_update_follow_up(test_chats_1):
+def test_create_follow_up_unauthorized(test_chats_1):
```
JnBrymn, line 44: 🥳 yay for all the new tests!

pennychat/views.py
```diff
@@ -16,8 +19,26 @@ class PennyChatViewSet(viewsets.ModelViewSet):
     serializer_class = PennyChatSerializer
 
     def perform_create(self, serializer):
-        # TODO add participants and make request user's UserProfile as the Organizer
-        serializer.save()
+        if not self.request.auth:
+            raise NotAuthenticated
+        chat = serializer.save()
+        profile = UserProfile.objects.get(user=self.request.user)
```
JnBrymn, line 25: this is a great example of "what on earth do we do?!" when there is more that one UserProfile. 🤔

JnBrymn:
Another thought - is it time to delete this: https://github.com/penny-university/penny_university/blob/5c7a630374de1953a2e27a8ae1475579d3833c11/users/models.py#L29-L30 and make sure that the user profile table is just named according to the django default?

pennychat/permissions.py
nchouard: You're right, I should add documentation.

nchouard:
Re: User table, I agree, somehow I just saw that. Going to try to remove

pennychat/views.py
nchouard: I decided to keep it as is. Since the rest framework generally uses permission classes, I'd like to stick to that paradigm, or use authentication decorators as necessary.

Issue #132, opened 2020-02-16
Remind attendees of details just before the chat


JnBrymn:
# TODO
- [x] Create reminder comment post
- [x] Create background job that does all the reminding for PennyChats that are coming in the next hour.
- [x] Make sure the reminder only reminds for a penny chat once
- [x] Update scheduled events in heroku
  * And make sure we're not using too many resources
- [x] Test
- [x] Cleanup (TODO!)
- [x] Deploy
- [x] Proof of Correctness ~ I tried `/penny chat` for a chat within 40 minutes and it reminded me.

JnBrymn:
# Old Description

### UPDATE
This ticket is no longer valid. I'm first doing https://github.com/penny-university/penny_university/issues/153 to move slack messaging into background tasks. Then I will reevaluate this.

### TODO
- [x] When user says they can attend, schedule reminder - make sure it is presented in their timezone
- [x] When user says the can no longer attend, delete reminder.
- [ ] When Chat date/time is changed, change the reminders for everyone.
- [ ] Test
  - [ ] Make sure timezones make sense!
- [ ] Show evidence of correctness

### NOTES
* https://api.slack.com/messaging/scheduling

### Considering Alternative Implementation
In order to complete the above we would have to add potentially LOTS more requests inside of the `submit_details_and_share` method in order to delete all the participants scheduledMessages and post new ones. (So `2*num_participants` more requests). This is the one that I'm already concerned will timeout.

So I'm looking into a combination of the [Heroku Scheduler](https://devcenter.heroku.com/articles/scheduler) and [Django Background Tasks](https://django-background-tasks.readthedocs.io/en/latest/).

PR #160, opened 2020-04-06
Remind users of PennyChats an hour before they occur.

JnBrymn:
# Context
We need to remind people to attend penny chat events.

# Approach
* I modified the existing method `tested penny_chat_details_template` to have variants for `PREVIEW`, `INVITE`, `UPDATE`, and `REMIND`.
* I added a new command that is intended to run background jobs periodically. This will include
  * The background tasks from my last PR (e.g. the long-running requests to slack that we can't run in the normal request cycle).
  * Periodically checking the database for "imminent penny chats" (upcoming within the hour) and remind their attendees of the chat.
  * In the future we can also remind users to submit followups and we can mark chats as `status=COMPLETE`
* `send_penny_chat_reminders` is the actually method that sends reminders

# Related
parent https://github.com/penny-university/penny_university/issues/132


bot/tasks/pennychat.py
```diff
@@ -73,59 +76,100 @@ def share_penny_chat_invitation(penny_chat_id):
     penny_chat_invitation.save()
 
 
-def datetime_template(penny_chat):
-    timestamp = int(penny_chat.date.astimezone(utc).timestamp())
-    date_text = f'<!date^{timestamp}^{{date_pretty}} at {{time}}|{penny_chat.date}>'
-    return date_text
+def send_penny_chat_reminders():
+    """This sends out reminders for any chat that is about to happen."""
+    slack_client = _get_slack_client()
```
nchouard, line 81: Why doesn't this have the `@background` decorator? I'm confused as to how this is processed.

penny_university/management/commands/background_tasks.py
```diff
@@ -0,0 +1,19 @@
+from background_task.management.commands.process_tasks import Command as ProcessTasks
+from background_task.utils import SignalManager
+from django.conf import settings
+
+from bot.tasks import send_penny_chat_reminders
+
+
+class Command(ProcessTasks):
+    def handle(self, *args, **options):
+        send_penny_chat_reminders()
```
nchouard, line 10: Oh, I missed it right here. I'm still a bit confused about what needs the background decorator and what doesn't.

bot/tasks/pennychat.py
JnBrymn: Since I _don't_ `@background` this function then (as you are thinking) it is NOT run in the background using that new framework. Rather I'm just sticking related functions into the same file.

penny_university/management/commands/background_tasks.py
JnBrymn: The previous `process_tasks` command (which is currently running in production) turns on the task queue that I installed. This command will replace it. It will effectively serve the same purpose, but before it start it will also quickly check the database for penny_chats that need reminders and fire them off. After that it will be in `process_tasks` mode. I'm using [heroku's scheduler](https://devcenter.heroku.com/articles/scheduler) to run this command every 10 minutes.

Eventually `send_penny_chat_reminders` will become time consuming enough that we'll have to split them out again... but that will be quite some time.

JnBrymn:
@nchouard you are right to be a little confused. I think the organization is a bit strange, but I didn't see an obvious correct way. If you see something better then consider branching off of my branch and make a PR to show me. Otherwise let me know how we can wrap up this review.

nchouard:
Your explanation makes sense! Thank you. Consider it 🚢 ed

PR #169, opened 2020-04-12
Authentication redux and middleware

CoreyAR:
# Description
Adds redux and middleware for basic authentication flows.
- Signing up
- Logging in
- Logging out
- Hydrating persisted session

User authentication is arbitrarily persisted with a cookie for 1 month

# Details

At the root of the application on mount an action is dispatched to start user hydration.
If a token is found, a thunk is sent to attempt to fetch the current user.
Other authentication flows are not implemented beyond the required redux /  middleware.


penny_university_frontend/src/actions/user.js
```diff
@@ -0,0 +1,69 @@
...
+
+export const checkAuth = () => ({
+  type: CHECK_AUTH,
+})
+
+export const fetchUser = () => ({
+  [CALL_API]: {
+    types: [FETCH_USER_REQUEST, FETCH_USER_SUCCESS, FETCH_USER_FAILURE],
+    endpoint: 'me/',
+    schema: Schemas.USER,
```
nchouard, line 34: Just making a note that we don't have this endpoint on the backend yet so I will create a task to do so.

PR #170, opened 2020-04-13
Extend Airbnb eslint config

CoreyAR:
# Description
Moves eslint config to .eslintrc and extends [airbnb config](https://github.com/airbnb/javascript) with minor customizations. A few rules are ignored with the intention of turning them back on when they are resolved.
The auto linter was also run on this PR and a few manual fixes were made such as updated keys to not use index, line spacing, and removing shadow vars.

# Details
### Turned off
- Require semi colons
- Prefer default export
### Temporarily set to warn
- no-shadow: Variables can't share a name with variables in the outer scope
- react/prop-types: Require Prop validation
- react/forbid-prop-types: Forbid vague prop types (any, array, object) 
- react/require-default-props: Optional props require a default
- react/default-props-match-prop-types: Default props must be appropriate type


dev/pre-commit
```diff
@@ -30,5 +30,15 @@ if __name__ == '__main__':
...
+        
+        print('\CHECKING ESLINT')
+        try:
+            subprocess.check_call('cd penny_university_frontend && yarn lint', shell=True)
+            print('\nPROCEEDING WITH COMMIT\n')
+        except subprocess.CalledProcessError as e:
+            print('\RUNNING ESLINT --FIX')
+            subprocess.check_call('cd penny_university_frontend && yarn lint --fix', shell=True)
+            frontend_errors = 1
+            print(NO_COMMIT_MSG)
```
JnBrymn, line 42: thanks

penny_university_frontend/src/middleware/api.js
```diff
@@ -92,34 +88,34 @@ export default store => next => action => {
...
   }
-  if (!types.every(type => typeof type === 'string')) {
+  if (!types.every((type) => typeof type === 'string')) {
     throw new Error('Expected action types to be strings.')
   }
 
-  const actionWith = data => {
-    const finalAction = Object.assign({}, action, data)
+  const actionWith = (data) => {
+    const finalAction = { ...action, ...data }
```
JnBrymn, line 103: TIL this destructuring pattern. Javascript has gone and gotten cool since I was an ember.js monkey.

CoreyAR: Yeah, I like how JS is evolving. Optional chaining is another cool feature we might want to take advantage of - https://github.com/tc39/proposal-optional-chaining

PR #178, opened 2020-04-17
Refactor buttons into reusable component

CoreyAR:
### Creates `<ButtonIcon/>` Component
All of the buttons had a similar layout and prop interface so I figured it made sense to compose them from a single base component. I liked how they were all exported individually so I kept that pattern to keep clean prop logic and avoid a boolean trap. Each button still keeps the same interface which passes the props to the underlying <IconButton />

The refactor was brought about so I could cleanly add styles to fix collapsing white space in the buttons
I also added defaultProps and propTypes to button components.


penny_university_frontend/src/components/buttons/index.jsx
```diff
@@ -0,0 +1,124 @@
+import React from 'react'
+import { FontAwesomeIcon } from '@fortawesome/react-fontawesome'
+import {
+  faPlus, faTimes, faPen, faSave, faHeart,
+} from '@fortawesome/free-solid-svg-icons'
+import classNames from 'classnames/dedupe'
```
CoreyAR, line 6: Seems like we have this library as a sub dependency. I worked on this along side #179 and it is explicitly added in that PR

PR #179, opened 2020-04-17
Add watch command; Move sass to frontend dir

CoreyAR:
This goals of this PR is to give us a more flexible and easier to use sass setup. 

- Move sass files to be a child of the frontend directory as  opposed to a sibling with the frontend and backend. I think this is preferable because the sass files are not relevant to the backend code.
- Adds `node-sass-chokidar` as a convenient method of setting up a watch command.
- Compiles scss files written in `/src`. This allows for styles to be written next to a component making it easier to understand how they are applied and possible to utilize directory structure for naming conventions to help debugging css if desired.
- Ignore CSS files. These are generated and can be part of the build process but don't need to get committed.


penny_university_frontend/package.json
```diff
@@ -7,12 +7,13 @@
     "@fortawesome/free-solid-svg-icons": "^5.11.2",
     "@fortawesome/react-fontawesome": "^0.1.7",
     "bootstrap": "^4.4.0",
+    "classnames": "^2.2.6",
```
CoreyAR, line 10: I worked on this alongside #180 This library is used in that PR

PR #180, opened 2020-04-17
Update path to settings file

CoreyAR:
Corrects the path to the settings file in the development README instructions.


CoreyAR:
I changed it because the current path is looking in for `dev.py` in `penny_university.slack`.  I don't see a `slack` dir in the project in the project root but we do have a `settings` dir with files for environment settings.

PR #181, opened 2020-04-17
Break out homepage scss and compile it to static dir

CoreyAR:
Nick pointed out that I inadvertently broke the compilation for the home page css.

I updated our css files to break out home page and app scss. The homepage doesn't have the extra app scss based on the committed  home page css file but it is possible that the app has extra home page css still.

While the home page is separate, I added an extra command to the compilation script that copies the home page css to the static dir.

### DIFF
--- a/penny_university_frontend/package.json
+++ b/penny_university_frontend/package.json
@@ -42,7 +42,7 @@
     "lint:fix": "yarn lint --fix",
     "lint:scss": "sass-lint -v $npm_package_config_src_css",
     "lint:scss:fix": "sass-lint-auto-fix",
-    "compile:css": "node-sass-chokidar --include-path ./src --include-path ../scss --include-path ./node_modules $npm_package_config_src_css -o $npm_package_config_dest_css",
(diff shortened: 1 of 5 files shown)


penny_university_frontend/package.json
```diff
@@ -42,7 +42,7 @@
     "lint:fix": "yarn lint --fix",
```
nchouard, line 42: Is it necessary to use `yarn` over `npm`?

penny_university_frontend/package.json
```diff
@@ -42,7 +42,7 @@
...
     "lint:scss": "sass-lint -v $npm_package_config_src_css",
     "lint:scss:fix": "sass-lint-auto-fix",
-    "compile:css": "node-sass-chokidar --include-path ./src --include-path ../scss --include-path ./node_modules $npm_package_config_src_css -o $npm_package_config_dest_css",
+    "compile:css": "node-sass-chokidar --include-path ./src --include-path ../scss --include-path ./node_modules $npm_package_config_src_css -o $npm_package_config_dest_css && cp ./home.css ../../penny_university/static/css/styles.css",
```
nchouard, line 45: When I ran this locally, I got this error: `cp: ./home.css: No such file or directory` Is this just me?

penny_university_frontend/package.json
CoreyAR: oh no, that is just what I have setup and use out of habit

PR #183, opened 2020-04-18
Add Signup and Login UI

CoreyAR:
# Description
Adds a Sign In and Register user Flow
User Enters an email
If they have an account prompt them for password
If they dont have an account, show them the signup form
Fetch the authed user with the provided token

[Login](https://share.clearsummit.com/xQuW9EwW)
[Signup w/ error examples](https://share.clearsummit.com/xQuW9EwW)

# Details
I updated the signup and registration endpoints to use `email` instead of `username`. I did this because login requires an email that is also used as a username and requires validation. Updated the key keeps it consistent.

Form validation is handled by bootstrap/ html5. It is a little wonky, for example 'email@domain' is considered a valid email.

The Error handling was lightly refactored to use a standard action. This is useful because we can see the full error object and respond appropriately to error codes without having to match strings.

The handling of the email model submission checks the returned promise directly. I don't think this is ideal but I wanted to make sure I was responding to the right error as out store doesn't support errors from multiple endpoints. I also think it would be useful to move this into middleware / sideffect and out of the view.
https://github.com/penny-university/penny_university/compare/issue/151/auth-flow-ui?expand=1#diff-a098844af8f689c99088fbd84db67a17R19-R30

There is also a small bug fix where on refresh there userProfile can get accessed before it exists.


penny_university_frontend/src/containers/AlertContainer.jsx
```diff
@@ -11,11 +12,16 @@ const AlertContainer = ({ error, dismissError }) => (
 
 const mapStateToProps = (state) => {
   const { error } = state
-  return { error }
+  return { error: error?.message }
```
JnBrymn, line 15: TIL `?.` - wow JS has gotten pretty cool since I've been away!

penny_university_frontend/src/tests/Auth.test.js
```diff
@@ -26,7 +26,7 @@ describe('auth flow', () => {
 
   it('should fetch user if token exists', () => {
     document.cookie = 'token=token;'
-    fetchMock.getOnce(`${baseUrl}me/`, {
+    fetchMock.getOnce(`${baseUrl}auth/user/`, {
```
JnBrymn, line 29: Should we be using `ApiRoutes` for these?

nchouard:
One more thing, which might have been a bug that already existed before this change but I'm only noticing it now. If you edit a follow up and then click to save the changes, it errors (which is of expected), but then the user attached to the follow up disappears.
edit
after save

CoreyAR:
The warnings in the console are related to PropTypes. It doesn't seem like there is a nice way to create a PropType that is required but also nullable so I left them. I can fix the number type error easily.  After digging into PropTypes some it really makes my miss complete static type checking.

I'll make the copy / color changes.

I tried making the email required  with html5 but it lost email validation. I'll do this in the formSubmit.

PR #209, opened 2020-05-07
Added using SLACK_INVITE_LINK to invite user

### LINKED ISSUE #39: Automatically add to slack rather than just telling us the email address.
Currently, if the user goes to pennyuniversity.co and clicks the Join button, their email will be sent to slack and then we will add them manually. Let's just have a machine do the whole process (though still notify admins in slack). And let's also post a notification in the #welcome-committee channel.

changhsinlee:
## What

* Changed the behavior of "Join Us on Slack" button
  - If `SLACK_INVITE_LINK` is not set, then use the current modal
  - If set, then take user to the link in a new tab when click on button
* This requires `SLACK_INVITE_LINK` to be set in Heroku Config Vars!!

## Why

Simplify the signup process for admins

## Reference

How to make a permanent Slack invite link: https://youtu.be/C5HAA8O_jZ8

Closes #39 

## Question

Not sure how to make the change in `index.html` template more concise without using the `{% if ... %}...{% endif %}` filter multiple times 🤔 suggestion welcome.

### DIFF
--- a/home/templates/home/index.html
+++ b/home/templates/home/index.html
@@ -25,12 +25,8 @@
                     <a href="https://twitter.com/officialpennyu" class="fa fa-twitter"></a>
                     <a href="https://www.instagram.com/officialpennyu" class="fa fa-instagram"></a>
                 </div>
-                <button type="button" class="ml-2 btn btn-outline-dark"
-                        {% if invite_link %}
-                        onclick=" window.open('{{ invite_link }}','_blank')"
-                        {% else %}
-                        data-toggle="modal" data-target="#formModal"
-                        {% endif %}>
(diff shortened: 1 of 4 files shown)


JnBrymn:
Oh holy cow! I never noticed that I could create an invite link like that. Thanks for making the video.

As for the implementation, your code obviously works, but I still stand by my previous comment that being dumped into slack directly is pretty jarring. If, on the other hand, you updated the modal text, then I think there would 1) be less duplicated logic and 2) the user would understand what it happening and we would get more people clicking through.

changhsinlee:
I'll modify the code and make an update later

changhsinlee:
Closing this PR, I'll make another one

PR #210, opened 2020-05-08
Replace Join Us modal to use permanent invitation link

### LINKED ISSUE #39: Automatically add to slack rather than just telling us the email address.

changhsinlee:
## What

* Changed the behavior of "Join Us on Slack" button
  - If `SLACK_INVITE_LINK` is not set, then keep the current modal
  - If set, then user sees the following modal where the button takes user directly to Slack sign-up page.
image

* This requires `SLACK_INVITE_LINK` to be set in Heroku Config Vars!!

## Why

Simplify the signup process for both users and admins

## Reference

How to make a permanent Slack invite link: https://youtu.be/C5HAA8O_jZ8

Feel free to test the new behavior here: https://my-fake-penny1.herokuapp.com/

Closes #39

### DIFF
--- a/home/templates/home/modal.html
+++ b/home/templates/home/modal.html
@@ -11,23 +11,18 @@
                 {% csrf_token %}
                 <div class="modal-body">
                     <p class="font-weight-light">
-                        Give us your email address and we will add you to our Slack group.
-                        You will also be able to add your own Penny Chat Reviews to our Google Forum.
+                        The Penny University community hangs out in Slack.
+                        Click the button below to go to our Slack sign-up page.
                     </p>
-                    {% for field in form %}
-                        <div class="form-group">
-                            {{ field.errors }}
-                            {{ field.label_tag }}
-                            {{ field }}
-                            {% if field.help_text %}
-                                <small class="form-text text-muted">{{ field.help_text|safe }}</small>
-                            {% endif %}
-                        </div>
-                    {% endfor %}
+                    <p>Not sure yet?
+                            <a href="https://groups.google.com/forum/#!forum/penny-university">
+                                Check out what we've been up to on our forum.
+                            </a>
+                        </p>
                 </div>
                 <div class="modal-footer">
                     <button type="button" class="btn btn-secondary" data-dismiss="modal">Close</button>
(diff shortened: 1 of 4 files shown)


home/views.py
```diff
@@ -12,24 +12,27 @@
...
-        if form.is_valid():
-            # process the data in form.cleaned_data as required
-            # send John and Nick a slack message with the person's information
-            email = form.cleaned_data["email"]
-            found_us = form.cleaned_data["how_did_you_find_us"]
-            message = f'Somebody just submitted a form!\nEmail: {email}\nHow They Found Us: {found_us}'
-            notify_admins(slack_client, message)
-            # redirect to a new URL:
-            return HttpResponseRedirect('/thank-you/')
+        if settings.SLACK_INVITE_LINK:
```
JnBrymn, line 15: Could you add a `notify_admins` with "Somebody clicked Join." here? Then we can compare this to the number of people actually joining and get a sense of how many people. This might get spammy, but it's not hard to change if it is.

home/templates/home/modal.html
```diff
@@ -9,6 +9,23 @@ <h5 class="modal-title" id="formModalLabel">Join Penny University</h5>
...
+                <div class="modal-body">
+                    <p class="font-weight-light">
+                        The Penny University community hangs out in Slack.
+                        Click the button below to go to our Slack sign-up page.
+                    </p>
+                    <p>Not sure yet?
+                            <a href="https://groups.google.com/forum/#!forum/penny-university">
+                                Check out what we've been up to on our forum.
+                            </a>
+                        </p>
```
JnBrymn, line 22: After looking at this in the context of the form, the "Not sure... Check out ... forum" text feels distracting from the goal of getting them into slack. Just delete it. (The forum will be gone soon anyway.)

changhsinlee:
@JnBrymn removed google forum link and added notified admin

When people click on button:
image

PR #212, opened 2020-05-09
Convert project to use Typescript

CoreyAR:
# Description
This PR adds Typescript to the entire project

# Context
I volunteered to spear head adding Typescript because I wanted to learn more about how it compares to Flow and because I believe static typing increases efficiency and protects against many bugs. 

This is a fairly large change set because it touches every file in the project. Thankfully github doesn't loose the change set when renaming files 🙏 

I reviewed this PR and left comments where I saw changes beyond the addition of types.

# Notes
I found typing our actions and store to be a little tricky so there are some ignored types that we can gradually improve on. The difficulty came from dispatching actions that could be StandardAction type objects or promises. Using the provided Redux generic types also proved to a little difficult


penny_university_frontend/src/components/modal/auth/email.tsx
```diff
@@ -2,29 +2,28 @@
...
         <Form onSubmit={(e) => {
           e.preventDefault()
-          userExists(email).then((action) => {
-            if (action.payload?.status === 400) {
-              modalDispatch.authSignup(email)
-            } else {
-              modalDispatch.authPassword(email)
-            }
-          })
+          userExists(email)
```
CoreyAR, line 26: This became obvious to me how to clean this up without catching the dispatch promise in the view. Handle in the middleware!

penny_university_frontend/src/containers/ChatsPage.tsx
```diff
@@ -1,42 +1,52 @@
...
+type ChatPageProps = StateProps & DispatchProps
+
+const ChatsPage = ({ filteredChats, nextPageUrl, loadChatsList }: ChatPageProps) => {
   useEffect(() => {
     loadChatsList('all')
   }, [loadChatsList])
   return (
     <div>
       <ChatList chats={filteredChats} />
-      <Button className="mb-3" onClick={(e) => loadChatsList('all', nextPageUrl, e)}>Load More</Button>
```
CoreyAR, line 16: We weren't using this extra argument that was getting passed

penny_university_frontend/src/actions/index.ts
```diff
@@ -76,10 +79,11 @@ const postFollowUp = (url, followUp) => ({
 })
 
 // These are all thunks. They return a call to dispatch with an action passed into it
-export const loadChatsList = (filter, nextPage) => (dispatch, getState) => {
+export const loadChatsList = (filter: 'all', nextPage?: string) => (dispatch: ThunkDispatch<{}, {}, AnyAction>, getState: () => RootState) => {
```
CoreyAR, line 82: I wonder if we need to curry here and if dispatching a returned action directly would be clearer

penny_university_frontend/src/index.jsx
```diff
@@ -1,27 +0,0 @@
...
-import rootReducer, { initialState } from './reducers'
-import api from './middleware/api'
-import user from './middleware/user'
-import './style.css'
-import App from './App'
-
-// Eventually we will want to move this into a DEV configuration
-const composeEnhancers = window.__REDUX_DEVTOOLS_EXTENSION_COMPOSE__ || compose;
-
-const store = createStore(rootReducer, initialState, composeEnhancers(
```
CoreyAR, line 16: I moved the creation of the store to the reducers index

penny_university_frontend/src/actions/index.ts
```diff
@@ -53,7 +56,7 @@ const fetchFollowUps = (chatId, nextPageUrl) => ({
   },
 })
 
-const putFollowUp = (url, followUp) => ({
+const putFollowUp = (url: string, followUp: FollowUp) => ({
```
JnBrymn, line 59: Where's a `FollowUp` defined?

CoreyAR: Because it was a type that would be used throughout the app I put it in a type declaration file so that it is available globally.

The type declaration file with `FollowUp` is at `reducers/types.d.ts`

PR #215, opened 2020-05-13
Refactor various user models/relationships

nchouard:
_opening as a draft for now because the Slack outage is preventing me from testing in my staging environment_

# Description
As we discussed in our meeting on 5/5, we want Penny Chats and Follow Ups to be related to `Users` rather than to what was previous called a `UserProfile`. To avoid confusion, `UserProfile` was also renamed to `SocialProfile`, since those entities will relate to Slack users or perhaps other services in the future. `PennyChatInvitation` was also renamed to `PennyChatSlackInvitation` for clarity.

## Details
Chats and Follow Ups now relate to Users rather than Slack profiles. All of the data can still be imported in from the google forum and slack. This screenshot shows an imported chat, and the second screenshot shows follow ups.
New user relation chats
New user relation follow ups

The `/users` endpoint now points to users and their chats, rather than social profiles, and a list of social profiles was added to the user serializer.
User detail
User chats


users/models.py
```diff
@@ -93,14 +93,14 @@ def update_social_profile_from_slack_user(slack_user):
             email=profile.email,
```
JnBrymn, line 93: nit: I think this is more understandable `first_name, last_name = "Tommy Lee Jones".split(maxsplit=1)` that get's you the same result

nchouard: I can't make it that clean unfortunately because there aren't always two values to unpack. I tried something else instead.

nchouard:
Big change after the previous work. I extended the AbstractUser so that we could have our own user class. Here is some proof of it working. Here is a user joined with a social profile after I created a pennychat in my staging environment (user did not previously exist):
```
         email          | slack_team_id |  real_name   |        username
------------------------+---------------+--------------+------------------------
 nick.chouard@gmail.com | TMTGU0YGJ     | Nick Chouard | nick.chouard@gmail.com
```

And here is the PennyChat joined with the user participant:
```
             title              |  description  |        username        | role
--------------------------------+---------------+------------------------+------
 Testing Penny Chat w/ New User | Does it work? | nick.chouard@gmail.com |   10
```

I played around a bunch in my staging environment, and everything seems to be working as expected.

Right before we merge, we will have to reset the database because we need to start from scratch with the new migrations. To do this, we need to manually run `heroku pg:reset -a penny-university`. Then, after pushing and building, our new migrations will run.

users/models.py
JnBrymn: boooooo ... good call, wasn't thinking of that

users/models.py
```diff
@@ -9,8 +9,13 @@
 from common.utils import pprint_obj
 
 
+class User(AbstractUser):
+    # Eventually, we will add fields here to extend this class
+    pass
```
JnBrymn, line 14: very cool

Issue #203, opened 2020-05-03
Create Integration Tests


JnBrymn:
# NEW GOAL
Make logger controlled by env var that will print out copy-pastable code for integration tests. When an endpoint is called, it will print out
```
client = APIClient()
response = client.post('/the/endpoint/', data={...})
assert response == 'whatever'
``` 

And if there are any clients to external services, then the clients can be decorated so that their input/output is printed too as
```
slack_client = mock.Mock()
slack_client.do_stuff.side_effect = [resp1, resp2,...]
...
assert slack_client.call_args = call('this')
```

# TODO
- [x] Figure out how to run background_task jobs synchronously ~ 84016e508f08
- [x] Make sure we can intercept messages to slack. ~ I made a slack client provider for this that we can easily mock
- [x] Make some actual tests.
  - /penny chat ~ done
  - ~onboarding~ ~ decided I was too tired
  - ~reminders~ ~ already tested sufficiently
- [x] Make some form of logging so that it's easier to make integration tests in the future. All you have to do is turn on a env var and then you can copy-paste logs as tests.
- [x] Document
- [x] Add tests to GitHub workflow
- [x] Make pytest run with the `ci` settings file and remove all the places where I used `backgrounded_task.now(*)` in the tests.
- [x] Make slack notifications for the production deploy.

JnBrymn:
# Original TODOs
- [ ] Create a quick-n-dirty flask app that records all its traffic and has a single defined endpoint that that can tell you if some call has been seen.
- [ ] Set etc. hosts on actions runner to redirect slack traffic to the flask app.
- [ ] Make sure that strategy _actually works_ in the github action
- [ ] Write tests to test that the proper slack endpoints are being called.

nchouard:
This ticket is done, we have a new one for the new integration tests.

PR #219, opened 2020-05-18
Integration tests and helper framework.

JnBrymn:
# What, why
* Added testing of /penny chat flow including summoning the modal, editing it, submitting it, and sending out invitations. because we needed integration tests to be able to deploy code more confidently.
* Added an "integration test logging" setup to log incoming web requests and outgoing client requests to slack. This was to make it possibly to copy-paste the logs as tests.
* Added docs about the integration testing.
* Updated pytest configuration to new, required settings.
* Removed `test-integration` section of github actions since the the new tests are run in the `test-backend` suite.

# Related 
Parent ticket https://github.com/penny-university/penny_university/issues/203


penny_university/settings/ci.py
```diff
@@ -1,4 +1,17 @@
+# from penny_university.settings.base import *
+#
+# # SECURITY WARNING: don't run with debug turned on in production!
+# DEBUG = True
+#
+# TASK_ALWAYS_EAGER = True
```
nchouard, line 6: Why are these commented out instead of deleted?

CoreyAR: I would expect that it would be always eager in CI for testing.

penny_university/middleware/integration_test_logging.py
```diff
@@ -0,0 +1,69 @@
+import logging
+from functools import wraps
+from unittest import mock
+
+from django.conf import settings
+from django.core.exceptions import MiddlewareNotUsed
+
+
+class IntegrationTestLogging:
```
CoreyAR, line 9: It would be cool to extend this to all requests and responses so we don't have to maintain mock data manually for frotend tests

penny_university/settings/base.py
```diff
@@ -60,7 +60,8 @@
     'django.contrib.auth.middleware.AuthenticationMiddleware',
     'django.contrib.messages.middleware.MessageMiddleware',
     'django.middleware.clickjacking.XFrameOptionsMiddleware',
-    'penny_university.middleware.DebugPassthrough',
+    'penny_university.middleware.debug_passthrough.DebugPassthrough',
+    'penny_university.middleware.integration_test_logging.IntegrationTestLogging',
```
CoreyAR, line 64: I think our logging is currently to standard out. There may be situations where we want to log to a file to commit for the frontend to use in tests.

penny_university/settings/ci.py
JnBrymn: silly me (they're undeleted/duplicated below this Corey)

penny_university/settings/base.py
JnBrymn: True and true. Feel free to upgrade us when you have need. The logging at this point is very naive.

penny_university/middleware/integration_test_logging.py
JnBrymn: I don't completely understand your statement here. As stated, this _is_ available for all requests and responses, no?

PR #223, opened 2020-05-22
Email Verification

### LINKED ISSUE #214: Require Email Validation
# Description
When users sign up to use our web app, we need to require some sort of email validation in order to verify that our users are who they say they are.

## Requirements
- [x] Set up an email service. I suggest we use sendgrid, since it will be free to send 40,000 emails in the first 30 days, and then 100/day after that. I have used it before and the API is very nice.
- [x] Send Users an email when they sign up for Penny U with a verification link.
- [x] Don't allow users to use the app until their email has been verified.

nchouard:
# Description
This change implements email verification. Users will be sent a verification email when they register. Users will not be able to log in to the web app until they verify.

closes #214 

### Flow
The flow will look as follows for the user:
- User makes a request to register and is sent a verification email.
- When the user follows the verification link, they are sent to a page in the web app that makes a request to the verify email url. The verification link contains their email and verification token.
- Once verified, they can be redirected to log in.

### Email
The verification email looks like this:
Verify
Originally, this email was sent to my spam. In order to prevent this, we will need to verify our domain, and I can replace the email that it is being sent from with an email from @pennyuniversity.org. This should prevent the emails from getting filtered to spam.


penny_university/settings/ci.py
```diff
@@ -1,7 +1,11 @@
...
+
 from penny_university.settings.base import *
 
 # SECURITY WARNING: don't run with debug turned on in production!
 DEBUG = True
 
 # If set to True, this ensures that any background tasks are run immediately rather than using django_background_tasks
 TASK_ALWAYS_EAGER = True
+
+FRONT_END_HOST = os.getenv('FRONT_END_HOST', 'http://localhost:3000')
```
nchouard, line 11: This is used in the `build_verification_url` method. I'm not sure if this is the best way to go about this, and would love some feedback.

users/models.py
```diff
@@ -143,3 +162,17 @@ def get_or_create_social_profile_from_slack_ids(slack_user_ids, slack_client=Non
...
+    Generates a url for the user to visit in the web app. If FRONT_END_HOST is not set, an exception will be raised.
+    :param email: A valid user email
+    :param token: Unique token to verify user email
+    """
+    base = settings.FRONT_END_HOST
+    query = {
+        'email': email,
+        'token': token,
+    }
+    return f'{base}/verify?{urlencode(query)}'
```
nchouard, line 178: @CoreyAR does this work for a URL to be directed to on the frontend? You can also just change it when you build out the functionality on the frontend.

penny_university/settings/ci.py
CoreyAR: Looks good to me

users/models.py
```diff
@@ -1,15 +1,34 @@
...
-    pass
+    is_verified = models.BooleanField(default=False)
+
+    def send_verification_email(self, token):
+        context = {
+            'verification_url': build_email_verification_url(self.email, token),
+            'first_name': self.first_name
+        }
+        text_email = render_to_string('users/verify_email.txt', context)
+        html_email = render_to_string('users/verify_email.html', context)
```
CoreyAR, line 24: I wonder if we shouldn't put all of this in sendgrid and use template vars to replace dynamic data like the link or names

users/models.py
CoreyAR: Works for me!

CoreyAR:
Looks good a couple thoughts 
When an unverified user tries to sign in should they be notified to verify when they enter their email or after they try to login with a password?

Is there an endpoint to resend the verify email?

Currently, it looks like we raise a login error and return the same error code as a failed login with incorrect credentials. I think we should differentiate these errors so we don't have to parse strings to show different UI for various errors.

I've been riffing off a pattern like this - 
```
{
	status_code: number, 
	description?: string, // developer message
	code?: string | number, // code to programmatically respond to the message 
	message?: {[field: string]: string, } | string // string or map (e.g. with form fields) with a UI message the frontend could show to a user
}
```
[Source](https://blog.restcase.com/rest-api-error-codes-101/)

penny_university/settings/ci.py
JnBrymn: I need this on the backend too https://github.com/penny-university/penny_university/pull/226/files#r429599376 - what's a good generic name for holding this `https://www.pennyuniversity.org` and this `http://localhost:3000`?

users/tests/api/test_user.py
```diff
@@ -3,46 +3,79 @@
...
+        username='other@test.com',
+        email='other@test.com',
+        password='password',
+    )
+    data = {
+        'token': verification_token.make_token(other_user),
+        'email': test_user.email
+    }
+    response = client.post('/api/auth/verify/', data=data, format='json')
+    assert response.status_code == 400
```
JnBrymn, line 54: great tests, helped me understand how everything works

users/tests/api/test_user.py
```diff
@@ -3,46 +3,79 @@
...
 
 
 @pytest.mark.django_db
-def test_checking_user_exists():
+def test_checking_user_exists(test_user):
```
JnBrymn, line 58: I'm of two minds w/ this type of test fixture. On the one hand it cleans up tests, but on the other hand it hides details that are useful for someone in the future that is unfamiliar with this test (in this case, the fact that the test_user has the email address `test@profile.com`). The fixture also couples tests together which can be frustrating when you need _kinda_ the fixture behavior but not quite the behavior. In this case it's pretty plain and simple, so maybe it doesn't matter. Decide what you think is best. I'm ok with either.

users/tokens.py
```diff
@@ -0,0 +1,10 @@
+from django.contrib.auth.tokens import PasswordResetTokenGenerator
+from django.utils import six
+
+
+class TokenGenerator(PasswordResetTokenGenerator):
+    def _make_hash_value(self, user, timestamp):
+        return six.text_type(user.pk) + six.text_type(timestamp) + six.text_type(user.is_verified)
```
JnBrymn, line 7: We shouldn't need any `six` stuff. This was used as a shim between python 2 and python 3. If you just cast to `str` then this should work fine ([see how `text_type` is defined here](https://docs.djangoproject.com/en/1.11/_modules/django/utils/six/)).

users/tokens.py
```diff
@@ -0,0 +1,10 @@
...
+
+
+verification_token = TokenGenerator()
```
JnBrymn, line 10: nit: This isn't a token, it's a token generator. I prefer `verification_token_generator`

penny_university/settings/ci.py
nchouard: @JnBrymn eventually they should point to the same place. For now I think it would be fine to just create another env var and then we can clean it up once we are serving the SPA from a django route.

penny_university/settings/base.py
```diff
@@ -172,5 +172,19 @@
 }
 
 
+# Twilio SendGrid
+EMAIL_HOST = 'smtp.sendgrid.net'
+EMAIL_PORT = 587
+EMAIL_USE_TLS = True
+EMAIL_HOST_USER = 'apikey'
+EMAIL_HOST_PASSWORD = os.environ.get('SENDGRID_API_KEY')
```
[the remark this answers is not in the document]
nchouard: It still works, it may just end up in a user's spam filter.

users/tests/api/test_user.py
nchouard: I found myself needing the fixture enough that it felt better to separate it. I think this way is best for this particular case, but I do see your point and it is something to keep in mind for the future.

users/tokens.py
nchouard: Thank you! I'll take a look at that. The examples I found for this probably came from an older django verison.

nchouard:
@CoreyAR @JnBrymn I updated the PR to address questions/concerns. Note the following:
* Added endpoint to resend verification emails.
* Updated UserExists view. It will check for verification and throw a 403 if the user isn't verified, or a 404 if the user doesn't exist.
* Bumped python version in docs because tests won't pass with <3.8.

users/tests/api/test_user.py
JnBrymn: 👍

PR #236, opened 2020-05-25
Edit User Details

### LINKED ISSUE #234: Edit User Detail endpoint
# Description
Create an endpoint to edit the User details.

## Details
The User Detail endpoint should allow PUT requests. Only the first and last name should be able to be changed for now.

nchouard:
# Description
Provides an endpoint to update the user's first and last name, and eventually more fields. closes #234 

## Details
* The user's first and last name can now be updated via a `PUT` or `PATCH` request to `users/{id}`. 
* The user serializer was split into a RegisterUserSerializer and a UserDetailSerializer so that the UserDetailSerializer would only return the relevant fields.

### DIFF
--- a/users/views.py
+++ b/users/views.py
@@ -1,5 +1,4 @@
 from rest_framework import (
-    mixins,
     generics,
     views,
 )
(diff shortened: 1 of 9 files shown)


users/views.py
```diff
@@ -90,13 +94,23 @@ def post(self, request, *args, **kwargs):
...
 
     def get(self, request, *args, **kwargs):
         return self.retrieve(request, *args, **kwargs)
 
+    @method_is_authenticated
+    @method_user_is_self
+    def put(self, request, *args, **kwargs):
+        return self.update(request, *args, **kwargs)
+
+    @method_is_authenticated
```
CoreyAR, line 109: What will be the different use cases for each of these?

CoreyAR:
Is there any reason for using both put and patch? Looks good otherwise

users/views.py
nchouard: The RetrieveUpdateAPIView implements both by default, so I'm overriding both as well in order to include the permissions decorators. I figure it gives us the flexibility of using either method in whatever way feels fitting on the frontend.

PR #257, opened 2020-06-10
Existing Users not Being Updated on Registration

nchouard:
# Description
We discovered that existing users (from a Penny chat participation) were not being updated during registration. This change fixes that, and also updates the invite email. To be from the Penny U domain.

## Details
Using my staging environment, I created a user and manually set their password to be unusable. I then registered the user, and received this email (note the updated address):
Updated Email

After verifying, I was able to properly sign in! 
verify
signed in

I also added a test for this particular case.


users/models.py
```diff
@@ -25,7 +25,7 @@ def send_verification_email(self, token):
         send_mail(
             'Welcome to Penny University | Verify Your Email',
             text_email,
-            'penny.university.mod@gmail.com',  # TODO: Replace email with one that has a Penny U domain
+            'invite@pennyuniversity.org',
```
JnBrymn, line 28: I just figured out that this formatted email works `'"Penny University" <invite@pennyuniversity.org>'`. Here's what that generates in gmail: 
image

You mind changing that?

nchouard: Oh sweet!

PR #263, opened 2020-06-17
Lint after typescript

CoreyAR:
# Description
Re-lints after typescript and changes the file extension from .js/.jsx to .ts/.tsx

Also adds linting to Frontend CI checks


nchouard:
All good from me, but looks like the frontend tests are failing.

PR #266, opened 2020-06-22
Reset Password

nchouard:
# Description
Creates a way for users to reset their password in our web app.

## Details
On the backend the PasswordReset serializer was overloaded from `dj-rest-auth`, and reset password email templates were added. The `dj-rest-auth` library takes care of the rest.

On the front end, users can now request a password reset and reset their password.
1. A user can click the "Forgot password" link in the log in modal, and they will be sent an email with a link to reset their password.
Forgot Password Link
Email Sent
Reset Email
2. The user can enter a new password.
Reset Password
3. When the user successfully resets their password, the will be given the option to log in, and then are redirected to the chats page.
Reset Success


common/templatetags/urls.py
```diff
@@ -0,0 +1,15 @@
...
+
+
+@register.simple_tag
+def build_url(base, endpoint, **kwargs):
+    """
+    Generates a url for the user to visit in the web app. If FRONT_END_HOST is not set, an exception will be raised.
+    :param base: The url base, i.e. https://pennyuniversity.org
+    :param endpoint: Unique token to verify user email
+    """
+    return f'{base}/{endpoint}?{urlencode(kwargs)}'
```
JnBrymn, line 15: If base is always the same then why make it an argument? Just us `FRONT_END_HOST` and raise if `FRONT_END_HOST` isn't defined.

penny_university_frontend/src/components/modal/auth/password.tsx
```diff
@@ -5,16 +5,17 @@ import { connect } from 'react-redux'
 import { ThunkDispatch } from 'redux-thunk'
 import { AnyAction } from 'redux'
 import {
-  Form, ModalHeader, ModalBody, Button,
+  Form, ModalHeader, ModalBody, ModalFooter, Button,
 } from 'reactstrap'
 import modalDispatch from '../dispatch'
-import { dispatchLogin } from '../../../actions/user'
+import {dispatchLogin, requestPasswordReset} from '../../../actions/user'
```
JnBrymn, line 11: nit, add preceeding and trailing ` `

users/models.py
```diff
@@ -18,7 +16,10 @@ class User(AbstractUser):
 
     def send_verification_email(self, token):
         context = {
-            'verification_url': build_email_verification_url(self.email, token),
+            'base': settings.FRONT_END_HOST,
+            'endpoint': 'verify',
+            'email': self.email,
+            'token': token,
```
JnBrymn, line 22: Personally I like building the verification_url outside of the template rather than inside the template. 1) Less logic inside of the template is always good. 2) The pieces of the url have to travel around together. Why not just make one thing, then url, which binds the pieces together and you thereafter only have to worry about a single object, not 4. Consider how complicated this would be if you had 3 urls. You'd have to ship around lots of variables and makes sure that the template put them together correct.

nchouard:
> `build_url` feels "clever" and I would prefer not ship with that (but you can if you want), other than that seems good. 🥂 

I thought it would create a nice abstraction so that we don't have to create different functions anytime we want to do this. I suppose it is a fine line between a good abstraction and a "clever" one. What don't you like about this particular approach?

common/templatetags/urls.py
nchouard: I figured it might not _always_ be the same.

users/models.py
nchouard: That's fair! How about I just change build_url to be a non-template tag? That way we still have a method that can be used in lots of different circumstances, but we don't have to pass as much around into the template.

JnBrymn: That sounds fine to me. In that case this code would look similar to where it started, but with a more general `build_url` method and we're back to having 2 context vars `verification_url` and `first_name`, right?

nchouard: Yup!

PR #280, opened 2020-07-03
Useful frontend error messages

### LINKED ISSUE #272: Bug in Forgotten Password.
If you sign on with the wrong password, the screen doesn't change but you get "An error occurred." message at the bottom. This is Game Over for the user because they can't do anything to log in.

image

nchouard:
# Description
Closes #272. Surfaces the actual error messages being returned from the API.

## Details
There can be multiple errors returned from the API, so they are mapped and shown on the screen in the alerts dialog. The dialog was given a z-index of 10000 in order to overlay everything on the page, including the modal. This way it is easily visible and can be cleared out.

While testing this, I noticed that we weren't validating passwords on registration, so I added that as well.

Here is an example of an incorrect password error:
Screen Shot 2020-07-02 at 7 14 40 PM


penny_university_frontend/src/middleware/api.ts
```diff
@@ -1,7 +1,8 @@
-import { MiddlewareAPI, Dispatch, Middleware, AnyAction } from "redux"
-import { camelizeKeys, decamelizeKeys } from 'humps'
+import {MiddlewareAPI, Dispatch, Middleware, AnyAction} from "redux"
+import {camelizeKeys, decamelizeKeys} from 'humps'
```
JnBrymn, line 2: You're deleting all the pre/post spaces. I don't really care, since it's syntactically equivalent, but it would be nice to settle on one or the other so that there are less red/green lines in diffs.

penny_university_frontend/src/middleware/api.ts
```diff
@@ -62,20 +63,42 @@ const api: Middleware<Dispatch> = (store: MiddlewareAPI) => (next: (action: AnyA
...
-      (error: { message: string, status: number }) => next({
-        type: failureType,
-        payload: { message: error.message || 'An error occurred.', status: error.status, meta },
-      }),
+      async (error: any) => {
+        let errorBody = null;
+        try {
+          await error.json().then((json: any) => {
+            errorBody = camelizeKeys(json)
+          }).catch()
```
JnBrymn, line 86: what does `catch` do here? if it does nothing then why have it?

penny_university_frontend/src/tests/components/Alerts.test.tsx
```diff
@@ -5,22 +5,25 @@ import { ErrorAlert } from '../../components/alerts'
 
 describe('error alert component', () => {
   it('renders correctly', () => {
-    const component = shallow(<ErrorAlert error="error" dismiss={jest.fn()} />)
+    const error = {error: "error"}
+    const component = shallow(<ErrorAlert error={error} dismiss={jest.fn()} />)
     expect(component).toMatchSnapshot()
```
JnBrymn, line 10: cool, hadn't noticed that this was a thing. How are snapshots updated?

penny_university_frontend/src/middleware/api.ts
nchouard: Yeah this is cause of using auto formatting and my IDE isn't set up the same. I'm gonna fix these.

penny_university_frontend/src/middleware/api.ts
nchouard: I was just following what was done with the non-error response to be honest.

penny_university_frontend/src/tests/components/Alerts.test.tsx
nchouard: They are updated if you delete them, and then run the test again.

Issue #277, opened 2020-07-01
On /chats/ page add the number of followups to each chat.


JnBrymn:
# Description
Right now the /chats/ page only lists the number of participants, but not the number of followups associated with each chat. This provides a better signal to let people know if it's worth clicking through.

## Details
**Requirements**
- [ ] Make sure the `/api/chats` endpoint returns a count of the FollowUps. [Here is a good StackOverflow answer to user as a guide.](https://stackoverflow.com/a/36697562)
- [ ] Write a test to check that the correct count is returned
- [ ] Display the follow up count on the ChatCard in the ChatList

PR #281, opened 2020-07-03
Chat Folloups Count and Filter

JnBrymn:
# what
Added a filter so that the /chats/ page would omit all chats with 0 followups AND added the followup count so that people could understand how "big" a Penny Chat was before clicking in. The /profile/<id> page is not filtered; it still shows chats with 0 followups.

image

# why
Because the /chats/ page can easily be filled with tests and cruft if we don't use some kind of filter.

# related
closes https://github.com/penny-university/penny_university/issues/278
closes https://github.com/penny-university/penny_university/issues/277


nchouard:
@JnBrymn just realized something else with this one. We want to see upcoming Penny Chats on the main page. Since upcoming chats don't have any follow ups, they won't show up. Can you adjust the filters so that we will get upcoming chats regardless of if they have follow ups or not?

PR #284, opened 2020-07-04
Post help message for all commands other than chat

### LINKED ISSUE #282: Post help message with any command sent to /penny other than chat
# Description
Currently if a command is sent that isn't `/penny chat` or `/penny help`, it doesn't work. We should return the help message any time a message is sent to `/penny` that isn't `/penny chat`.

nchouard:
# Description
Closes #282. Posts the help message in response to any command other than `/penny chat`, including empty commands like `/penny `

## Details
command
response


nchouard:
I think it is a permissions thing; just like the bot can't read from private channels, it can't write to them either.

JnBrymn:
Landable, right?

Issue #73, opened 2019-11-24
Welcome Back Event


JnBrymn:
# TODO
- [x] Email/slack all members that have been active in the past (below) a link to their "my chats" page.
  - [x] Write text for email. ~ https://github.com/penny-university/penny_university/blob/1d848192086c34dfbfd4f1ee3ef3b0da933d0838/users/templates/users/welcome_back_email.txt
  - [x] Create HTML template. (Probably using https://beefree.io/bee-pro/)
  - [x] Create manage command to gather all web app Users as well as a list of emails for the forum users and send an email. ~ just have to add templating to the HTML version
  - [x] Get reviewed, and then send to slack and forum users.
- [x] Help Nick update FAQ
- [x] Post fanfare in Slack.
- [x] Set up a few different large-group penny chats to let people know what's happening. ~ https://github.com/penny-university/penny_university/issues/73#issuecomment-654574757
- [x] Video walkthrough of features for the FAQ and drop that into slack with a TL;DR "use `/penny chat`" ~ https://penny-university.slack.com/archives/C41G02RK4/p1596637191220200

JnBrymn:
Outline for Penny Chat classes.
* Intro: 15 minutes to talk about what's new at Penny U followed by 15 minutes of questions and brainstorming.
* Context:
  * Penny U is _your_ social learning community and we've been looking for ways to build a platform to help bolster the good culture that we've seen emerge in the community.
  * The behavior that we've seen has been 
    * Organizing in Penny Chats - scheduling, inviting, and meeting.
    * Writing Penny Chat followups
  * Historically the "platform" has been Slack, Zoom, and the Google Forum.
  * But these tools have lead to difficulties:
    * Coordinating people with tons of messages is inconvenient.
    * Writing a follow-up the the Google forum is awkward.
* So we're building software to resolve these problems.
  * `/penny chat` to create a chat and invite channels or individuals (make sure to include video meeting link)
  * Get notified as people accept or decline invitations.
  * Everyone that RSVPed gets a slack reminder just an hour before the notification.
  * At the end of the meeting there is a slack reminder to write a followup. (Tell "how Penny U says thanks" story.)
  * And then /profile/X becomes a trophy room of all the interesting conversations you've had, things you've learned, and people you've spoken with.
  * Finally /chats/ is a good place to see all the interesting things that have happened in the past and upcoming events.
* In the future: 
  * scheduling
  * engagement (find an upcoming chat for you)
  * search
  * calendar?
[...]

PR #287, opened 2020-07-12
Send Welcomeback email.

JnBrymn:
I build an email to welcome people back. There are three variants as seen here https://github.com/penny-university/penny_university/issues/73#issuecomment-657107562

It emails all current users of slack and the old forum. (If there is overlap in these groups, we dedupe them so that everyone receives just one email.)

I need to sign up for the Sendgrid demo so that we send all ~600 emails at once because the current send limit is 100 emails per day.

Review this for:
1. 👍 for the code
2. 👍 sending an email to 600 people! 😨


JnBrymn:
Good idea. I added my email address if they want to contact us. I would have added yours too, but it seemed like a clearer message to just include a single email address. (🤔 though in retrospect, maybe I should have just made the message from you! 😛 )

PR #317, opened 2020-08-04
Merge Users

nchouard:
# Description
We have a couple of users who have two different emails for Penny U. They could have Follow Ups from the forum that are not associated with the account they've created for our platform. This changes introduces a management function that merges two users based on their emails.

## Details
To show proof of completeness, I created a user named Mr Merge. He had a Social Profile and created a Penny Chat and Follow Up.
before
Running this command, `python manage.py merge_users --from-email haxiwe1494@cartmails.com --to-email nick.chouard@gmail.com`, produces this output:
```
WARNING: SLACK_INVITE_LINK is None
Participation in "A Chat by Mr Merge" merged to nick.chouard@gmail.com
Follow Up from "A Chat by Mr Merge" merged to nick.chouard@gmail.com
Profile Mr Merge merged to nick.chouard@gmail.com
THIS IS A DRY RUN ONLY - NOT COMMITTING
Run with --live-run to actually commit.
```
After running the command with the `--live-run` flag, this is the output:
after
All of Mr. Merge's assets now belong to Nick Chouard.

### DIFF
--- a/docs/product_catalog.md
+++ b/docs/product_catalog.md
@@ -18,6 +18,7 @@ The goal of this document is to enumerate all the pieces so that we don't lose t
 * Internal API (see api directory)
 * Importing data from forum (pennychat/management/commands/import_google_forum.py)
 * Importing users from Slack (users/management/commands/import_users_from_slack.py)
+* Email Penny University members (users/management/commands/email_members.py)
 * Merge two users using their email (users/management/commands/merge_users.py)
 
 ## Other
--- a/docs/product_catalog.md
+++ b/docs/product_catalog.md
@@ -18,6 +18,7 @@ The goal of this document is to enumerate all the pieces so that we don't lose t
 * Internal API (see api directory)
 * Importing data from forum (pennychat/management/commands/import_google_forum.py)
 * Importing users from Slack (users/management/commands/import_users_from_slack.py)
+* Merge two users using their email (users/management/commands/merge_users.py)
 
 ## Other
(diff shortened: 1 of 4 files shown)


docs/product_catalog.md
```diff
@@ -18,6 +18,7 @@ The goal of this document is to enumerate all the pieces so that we don't lose t
 * Internal API (see api directory)
 * Importing data from forum (pennychat/management/commands/import_google_forum.py)
 * Importing users from Slack (users/management/commands/import_users_from_slack.py)
+* Merge two users using their email (users/management/commands/merge_users.py)
```
JnBrymn, line 21: nice! I forgot that this was here - do you mind adding users/management/commands/email_members.py ?

users/management/commands/merge_users.py
```diff
@@ -0,0 +1,51 @@
...
+            '--live-run',
+            dest='live_run',
+            action='store_true',
+            help='opposite of dry run - will actually write data to the database',
+            required=False,
+        )
+        parser.add_argument(
+            '--from-email',
+            dest='from_email',
+            help='The email of the user you want to merge into the other',
```
JnBrymn, line 25: Should we delete the old user? If so add "this user will be deleted"

nchouard: I think we should. I'll add that message.

docs/product_catalog.md
nchouard: Yep!

PR #320, opened 2020-08-05
Delete Follow Ups

### LINKED ISSUE #319: Ability to delete Followups
# Description
Users should be able to delete their follow ups.

## Details
**Requirements**
- [ ] From the frontend of our app, users should be able to delete follow ups.

nchouard:
# Description
Adds the ability for users to delete their own follow ups from the ChatDetail page. Closes #319 

## Details
The options menu was updated to include a delete button. The options menu was also changed from a popover to a reactstrap Dropdown, because the popover was causing some issues with the modal.
delete menu

When the delete option is clicked, a confirmation modal is opened.
delete modal

After clicking confirm, the follow up is deleted.
deleted

Right now, the way the way that the pagination reducer handles the deletion is not very elegant:
```javascript
else if (deleteTypes.includes(action.type)) {
      if (action.type === ChatActions.DELETE_FOLLOW_UP_SUCCESS) {
        const index = state.ids.indexOf(action.payload?.meta?.followUpID.toString())
        const newState = { ...state }
        newState.ids.splice(index, 1)
        return {
          ...newState,
          isFetching: false,
        }
     }
}
```
Up for suggestions on this.


penny_university_frontend/src/reducers/paginate.ts
```diff
@@ -23,27 +24,29 @@ const union = (a: Iterable<any>, b: Iterable<any>) => (
...
-  if (!types.every((t) => Array.isArray(t))) {
-    throw new Error('Expected types to be strings.')
-  }
+  {
+    types: {
+      requestTypes: Array<string>, successTypes: Array<string>, failureTypes: Array<string>, deleteTypes: Array<string>
+    },
+    mapActionToKey: (action: AnyAction) => string | undefined
+  }) => {
   if (typeof mapActionToKey !== 'function') {
```
JnBrymn, line 33: is this check still necessary? I would figure that Typescript would enforce this and compile error if it couldn't

penny_university_frontend/src/components/follow-ups/FollowUpCard.tsx
```diff
@@ -19,22 +25,34 @@ type FollowUpCard = {
...
-          type="Follow Up"
+        <DropdownItem
+          className="btn btn-link"
           onClick={editOnPress}
           key={`edit-followup-${id}`}
-          color="link"
-        />,
+        >
+          <FontAwesomeIcon icon={faPen} />
+          {' '}
```
coreyar, line 44: I don't think we should keep carrying this pattern foward. I imagine we will hit a bug where the browser collapses white space. A better solution is to use css for spacing.

penny_university_frontend/src/components/modal/confirmation/deleteFollowUp.tsx
```diff
@@ -0,0 +1,46 @@
...
+const DeleteFollowUpModal = ({ deleteFollowUp, followUpID, chatID }: DeleteFollowUpModalProps) => {
+  const deleteOnPress = () => {
+    deleteFollowUp(followUpID, chatID)
+    modalDispatch.close()
+  }
+  return (
+    <>
+      <ModalHeader toggle={modalDispatch.close}>Delete Follow Up</ModalHeader>
+      <ModalBody>
+        Are you sure you want to delete this Follow Up? This action cannot be undone.
```
coreyar, line 26: nit: We might want to think about putting our strings in a file in the event we want to add translations later

penny_university_frontend/src/reducers/index.ts
```diff
@@ -76,19 +81,21 @@ const errorReducer = (state = { status: NaN, body: null }, action: AnyAction): E
 const pagination = combineReducers({
   chatsByFilter: paginate({
     mapActionToKey: (action?: AnyAction) => action?.payload?.meta?.userID || 'all',
-    types: [
-      [ChatActions.CHATS_LIST_REQUEST],
-      [ChatActions.CHATS_LIST_SUCCESS],
-      [ChatActions.CHATS_LIST_FAILURE],
-    ],
+    types: {
```
coreyar, line 84: Nice

coreyar:
Looks good. Only comments I have are thoughts about maintainability.

penny_university_frontend/src/components/follow-ups/FollowUpCard.tsx
nchouard: I decided to replace the spaces with a `mr-2` utility class on the icon elements.

Issue #321, opened 2020-08-07
Privacy Functionality 


deadbender:
## Description

**Private Chat**
- visible on participant's profile, if the participant is the logged in user
- not visible on /chats to any users 
- if user accesses chat URL they do not have access to, send to 403

**Public Chat**
- visible to all signed in and not signed in users on profiles or /chats

**Etc**
- users should be able to click a checkbox in the `/penny chat` form to set private or public status
  - should eventually be editable from web app as well

## TODO
- [x] ~~Add `visibility` field to `PennyChat` in pennychat/models.py~~
  ~~* type `IntegerField`~~
  ~~* `choices` should be `PRIVATE = 10` and `PUBLIC = 20` (maybe there will be other types of visibility later)~~
  ~~* Update the db with `./manage.py makemigrations` then `./manage.py migrate`~~
- [x] ~~Add `public_or_participated` method to `PennyChat` in pennychat/models.py~~
  ~~* It will have an argument for `user`~~
  ~~* It will return a queryset of all Penny Chats that are `visibility = PUBLIC` or something attended by the `user`. This query will be a pain in the butt to write.~~
- [x] ~~In `PennyChatViewSet` in pennychat/views.py figure out how to use that filter to narrow down the list of penny chats~~
~~- [x] Within `bot.processors.pennychat.PennyChatBotModule.create_penny_chat` add a Private checkbox.~~
- [ ] Private Checkbox Styling
- [x] ~~Within `bot.processors.pennychat.PennyChatBotModule.submit_details_and_share` save the Private checkbox to the  `visibility` field of the PennyChatInvitation.~~

## TESTS

- [ ] Fix original tests with issues 

- [ ] Test to make sure `visibility` exists in `PennyChat` model
- [ ] Test to make sure private chats aren't visible to non participating users
- [ ] Test to make sure that participated chats are visible to participated users, even when private

PR #322, opened 2020-08-08
Private Chat vs. Public Chat

deadbender:
Adds in the functionality for Penny Chats to be marked as public or private on creations and for private chats to only be seen by their participants.

**To-Do**
- [x] Chat Visibility should grab the actual value, not use default, on edit details


deadbender:
Linking to #321

bot/processors/pennychat.py
```diff
@@ -142,6 +143,34 @@ def penny_chat_details_modal(penny_chat_invitation):
...
+                            },
+                            "value": "private",
+                        },
+                        {
+                            "text": {
+                                "type": "plain_text",
+                                "text": "public chat?",
+                            },
+                            "value": "public?",
+                        },
```
nchouard, line 166: Because of the options we currently have, I don't think you should have two checkboxes here. What would happen if the user selected both? There should either be one checkbox for private (if it's checked, it is a private chat, otherwise public), or a set of radio buttons so that only one option can be selected.

deadbender: The issue is that with a checkbox field, Slack requires an input to be selected and a value returned. I haven't seen a way to get around that but I'll look a little more. All I can say right now is if I stuck with only a private checkbox, for a public chat it would never allow the user to press submit. 

Radio buttons is a good idea, looking into the documentation for that.

nchouard: Ok good to know. I did have a feeling there might be some slack weirdness behind it :P.

JnBrymn: Agreed w/ Nick. Somehow we've got to get away from multiple options. It really needs to be _a_ checkbox. There's got to be some way to work around this.

deadbender: https://api.slack.com/reference/block-kit/block-elements#checkboxes Here is the docs on the checkbox, I don't see any field that'll let me disable required for the section :thinking: Let me know if there's something I'm missing. I need to clean that field up and make everything capitalized anyways.

nchouard: Why not radio buttons with Public as the default? https://api.slack.com/reference/block-kit/block-elements#radio. Not as clean as a single check box, but much better!

[You can also use this to see an example](https://app.slack.com/block-kit-builder/T0JC6MASV#%7B%22blocks%22:%5B%7B%22type%22:%22section%22,%22text%22:%7B%22type%22:%22plain_text%22,%22text%22:%22Check%20out%20these%20rad%20radio%20buttons%22%7D,%22accessory%22:%7B%22type%22:%22radio_buttons%22,%22action_id%22:%22this_is_an_action_id%22,%22initial_option%22:%7B%22value%22:%22A1%22,%22text%22:%7B%22type%22:%22plain_text%22,%22text%22:%22Radio%201%22%7D%7D,%22options%22:%5B%7B%22value%22:%22A1%22,%22text%22:%7B%22type%22:%22plain_text%22,%22text%22:%22Radio%201%22%7D%7D,%7B%22value%22:%22A2%22,%22text%22:%7B%22type%22:%22plain_text%22,%22text%22:%22Radio%202%22%7D%7D%5D%7D%7D%5D,%22type%22:%22home%22%7D), and click the Actions tab to see what info you would get back. Very handy!

nchouard: @JnBrymn 

> There's got to be some way to work around this.

You forget that we are working with the Slack API ;)

bot/processors/pennychat.py
```diff
@@ -145,22 +145,26 @@ def penny_chat_details_modal(penny_chat_invitation):
             },
             {
                 'block_id': 'penny_chat_visibility',
-                'type': 'input',
+                'type': 'static_select',
+                "placeholder": {
+                    "type": "plain_text",
+                    "text": "Select visibility"
+                },
```
JnBrymn, line 152: can we made a default option be Public? we'll definitely want the

bot/processors/pennychat.py
```diff
@@ -142,6 +150,48 @@ def penny_chat_details_modal(penny_chat_invitation):
...
+            },
+            {
+                'block_id': 'penny_chat_visibility',
+                'type': 'section',
+                'text': {
+                    'type': 'mrkdwn',
+                    'text': '*Chat Visibility*',
+                },
+                'accessory': {
+                    'action_id': 'penny_chat_visibility',
```
nchouard, line 164: Use PENNY_CHAT_VISIBILITY variable

bot/processors/pennychat.py
```diff
@@ -142,6 +150,48 @@ def penny_chat_details_modal(penny_chat_invitation):
...
+                        'type': 'plain_text',
+                        'text': 'Select visibility'
+                    },
+                    'options': [
+                        {
+                            'text': {
+                                'type': 'plain_text',
+                                'text': 'Private Chat',
+                            },
+                            'value': '20',
```
nchouard, line 176: Can these be ints instead of strings?

bot/processors/pennychat.py
```diff
@@ -142,6 +150,48 @@ def penny_chat_details_modal(penny_chat_invitation):
...
+                        },
+                        {
+                            'text': {
+                                'type': 'plain_text',
+                                'text': 'Public Chat',
+                            },
+                            'value': '10',
```
nchouard, line 183: Should be int

bot/processors/pennychat.py
```diff
@@ -34,9 +34,11 @@
 PENNY_CHAT_SHARE = 'penny_chat_share'
 PENNY_CHAT_CAN_ATTEND = 'penny_chat_can_attend'
 PENNY_CHAT_CAN_NOT_ATTEND = 'penny_chat_can_not_attend'
+PENNY_CHAT_VISIBILITY = 'penny_chat_visibility'
 
 PENNY_CHAT_ID = 'penny_chat_id'
 
+global penny_chat_visibility_name
```
nchouard, line 41: Why do we need this `global` variable?

bot/processors/pennychat.py
deadbender: For some reason (I believe it's just the way Slack API processes the JSON), using ints causes a syntax error and you can't dispatch the modal.

bot/processors/pennychat.py
deadbender: Oh yeah, not sure why I made that a global variable, that can be updated. I think my initial thought process was using it later down the line would cause an issue but I somehow just realized every use of it is in the same function 😄

bot/tests/processors/test_pennychat.py
```diff
@@ -138,6 +138,16 @@ def id_mock(user_id, slack_client=None, ignore_user_not_found=True):
                         'penny_chat_description': {
                             'value': 'new_description',
                         }
+                    },
+                    'penny_chat_visibility': {
+                        'private_chat_select': {
+                            'type': 'checkboxes',
```
nchouard, line 144: Why `'type': 'checkboxes'`?

bot/processors/pennychat.py
```diff
@@ -286,6 +336,18 @@ def channel_select(self, event):
...
+    @has_action_id(PENNY_CHAT_VISIBILITY)
+    def visibility_select(self, event):
+        selected_visibility = event['actions'][0]['selected_option']['value']
+        penny_chat_invitation = PennyChatSlackInvitation.objects.get(view=event['view']['id'])
+        if selected_visibility == '10':
+            penny_chat_invitation.visibility = 10
+            penny_chat_invitation.save()
+        if selected_visibility == '20':
+            penny_chat_invitation.visibility = 20
+            penny_chat_invitation.save()
```
nchouard, line 349: Since Slack won't allow you to use integers, could convert this to `penny_chat.visibility = int(selected_visibilty)` and eliminate the conditionals. Not a big deal though.

JnBrymn: I agree.

bot/tests/processors/test_pennychat.py
JnBrymn: I doubt that this code change is needed. Delete it and I bet the tests still pass because the final approach you chose doesn't even use this.

JnBrymn:
@deadbender [here's an example of using FactoryBoy to make fake objects](https://github.com/penny-university/penny_university/blob/7f1ae7b7c22c866dee1d865033d8e22ca7722020/users/management/commands/tests/test_notify_users_about_activity.py#L23). It's a good way to see how our models are wired together. Copy paste and modify some of this to the top of the tests I'm asking you to make and get rid of the `test_chats_1` fixture in the call arguments.

deadbender:
aha! i found [an example of how to appear to the API as an authenticated user](https://github.com/penny-university/penny_university/blob/dd27e86253a60f523da371b0398220b4773b241b/pennychat/tests/api/test_follow_up.py#L24)! this *should* be simple enough now 😄

deadbender:
Late night work but everything is almost done, only issue is that one specific test (seen above in checks) is failing on Github Actions but not my local machine.

bot/processors/pennychat.py
```diff
@@ -65,6 +66,12 @@ def penny_chat_details_modal(penny_chat_invitation):
     if penny_chat_invitation and len(penny_chat_invitation.channels) > 0:
         channels = comma_split(penny_chat_invitation.channels)
 
+    # giving a title to the visibility to make it dynamic
+    if penny_chat_invitation.visibility == 10:
```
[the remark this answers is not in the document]
JnBrymn: ☝️ do it! it makes the code easier to understand and less prone to bugs

bot/tests/processors/test_pennychat.py
```diff
@@ -32,6 +32,51 @@ def create_penny_chat():
     return chat.id
 
 
+def create_penny_chat__private():
```
JnBrymn, line 35: remove this function and instead create a new `PennyChatSlackInvitationFactory` with all the other factories that defaults to PUBLIC, then when you need this in the test do `penny_chat = PennyChatSlackInvitationFactory(visibility=PennyChat.PUBLIC)` and you can get the id as `penny_chat.id`

pennychat/tests/api/test_penny_chat.py
```diff
@@ -1,14 +1,47 @@
...
+
+    Participant.objects.create(user=user3, penny_chat=chat_private_2, role=Participant.ORGANIZER)
+    Participant.objects.create(user=user1, penny_chat=chat_private_2, role=Participant.ATTENDEE)
+
+    for chat in chats:
+        FollowUpFactory(penny_chat=chat, user=user1, date=within_range_date)
+    return {
+        'penny_chats': [chat_private_1, chat_private_2],
+        'users': [user1, user2, user3],
+        'social_profiles': [soc_prof1, soc_prof2, soc_prof3]
```
JnBrymn, line 42: Looks like you don't need social profiles for these tests. Remove 'em!

pennychat/tests/api/test_penny_chat.py
```diff
@@ -38,13 +77,34 @@ def test_penny_chat_participants_list(test_chats_1):
         assert int(user_id) in [participant['user']['id'] for participant in chat['participants']]
```
JnBrymn, line 77: This test proves that if a chat is returned, then it is a chat that they participated in. However it doesn't prove the corollary, that if they participated in a chat it would be returned EVEN if it's private.

I would therefore add something like this:
```
returned_chat_ids = set([chat['id'] for chat in chats])   # see https://book.pythontips.com/en/latest/comprehensions.html to understand this syntax
assert chat_private_1.id in returned_chat_ids, 'user organized this chat but it was not returned'
assert chat_private_2.id in returned_chat_ids, 'user attended this chat but it was not returned'
```

pennychat/tests/api/test_penny_chat.py
```diff
@@ -38,13 +77,34 @@ def test_penny_chat_participants_list(test_chats_1):
...
+@pytest.mark.django_db
+def test_penny_chat_participants_list__other_content():
+    objects = setup_test_chats()
+    chat_private_1, chat_private_2 = objects['penny_chats']
+    user1, user2, user3 = objects['users']
+    soc_prof1, soc_prof2, soc_prof3 = objects['social_profiles']
+    client = APIClient()
+    token = Token.objects.create(user=user3)
+    client.credentials(HTTP_AUTHORIZATION='Token ' + token.key)
+    user_id = user2.id
```
JnBrymn, line 89: nit pick - here in in the previous test I wouldn't create a `user_id` variable, it's actually easier to read `user2.id` and understand what's going on

pennychat/tests/api/test_penny_chat.py
```diff
@@ -38,13 +77,34 @@ def test_penny_chat_participants_list(test_chats_1):
...
+    response = client.get(f'/api/chats/?participants__user_id={user_id}')
+    assert response.status_code == 200
+    assert response.data['count'] == 0
+    chats = response.data['results']
+    for chat in chats:
+        assert int(user_id) in [participant['user']['id'] for participant in chat['participants']]
```
JnBrymn, line 95: Similar last test, you need extra test to prove that things that should be present are present and things that should not be present are not.

These 3 test cases: If I am looking at your profile
* I should see any public chat even if I didn't participate (this means you need to create a third public chat).
* I should see any private chat that you had where I was a participant.
* I should not see any private that that you participated in.

One way to make this test easier to compare to the last one is to look at the same profile page for both:

```
client.get(f'/api/chats/?participants__user_id={user1.id}')
```

And then it will be obvious what the assertions mean.

pennychat/tests/api/test_penny_chat.py
```diff
@@ -38,13 +77,34 @@ def test_penny_chat_participants_list(test_chats_1):
...
+
+
 @pytest.mark.django_db
 def test_penny_chat__upcoming_or_popular(test_chats_2):
```
JnBrymn, line 99: Similar for this test:

As a logged in user who looks at the /chats/ endpoint:
* I should be able to see all public chats
* I should be able to see all private chats where I was a participant
* I should NOT be able to see any private chats where I was not a participant

PR #329, opened 2020-08-18
Send User Notifications of Recent Followups

JnBrymn:


common/tests/fakes.py
```diff
@@ -0,0 +1,57 @@
...
+
+class UserFactory(DjangoModelFactory):
+    class Meta:
+        model = User
+
+    username = LazyAttribute(lambda u: u.email)
+    email = Faker('email')
+    password = 'password'
+    first_name = Faker('first_name')
+    last_name = Faker('last_name')
```
nchouard, line 22: Love the factory stuff

penny_university/settings/base.py
```diff
@@ -169,6 +169,8 @@
 # Put product specific configuration below this.
 ############################################
 
+PENNYU_DOMAIN = 'pennyuniversity.org'
+
```
nchouard, line 173: Why is this necessary? Why not use FRONT_END_HOST?

users/management/commands/notify_users_about_activity.py
```diff
@@ -0,0 +1,339 @@
...
+        }
+        if not people:
+            continue
+        people_string = get_people_string(people)
+        date_string = (
+            f'<!date^{int(penny_chat["date"].timestamp())}^{{date_short}}|{penny_chat["date"].strftime("%b %d, %Y")}>'
+        )
+        chat_lines.append(
+            f'Your {date_string} chat _"{penny_chat["title"]}"_ has follow-ups from {people_string}. '
+            f'*<http://{settings.PENNYU_DOMAIN}/chats/{penny_chat["id"]}|Read them here!>*'
```
nchouard, line 235: As with my comment above, could be replaced with FRONT_END_HOST

nchouard: Also, I wonder if a button would be better UX here. Do you have screen shots of what this looks like?

users/management/commands/notify_users_about_activity.py
```diff
@@ -0,0 +1,339 @@
...
+
+def get_people_string(people):
+    """Returns a pretty printed string for the list of people.
+
+    :param people: Assumes a dictionary of first_name-to-user_id
+    :return: String combining all the users
+    """
+    people_strings = []
+    for first_name, user_id in people.items():
+        people_strings.append(f'<http://{settings.PENNYU_DOMAIN}/profile/{user_id}|{first_name}>')
```
nchouard, line 259: Same comment

penny_university/settings/base.py
JnBrymn: done

JnBrymn:
Here's text:
image

if someone has lots of followups:

image


Here are the corresponding block versions:
image

and 

image

users/management/commands/notify_users_about_activity.py
JnBrymn: I agree. Changed in 6985f40

PR #333, opened 2020-08-22
Updated Interest Survey

nchouard:
# Description
Updates the interest survey. The interest survey has been shortened, and uses the modal API instead of the outdated dialog API. Closes #333 

## Details
The message for users to add their interests looks like this:
greeting

The new form has been simplified:
Form

The form still sends a message to the #general channel to let people know about this user's interests.
Channel message


nchouard:
Totally forgot about that. I'm gonna do that.

PR #339, opened 2020-08-26
[Issue-316] Revert auto save

coreyar:
Sending the chat when an un authed user needs to log in seemed like a good idea at first but it ended up introducing bugs and the implementation was unsatisfactory.

This PR reverts those changes in favor of a storing unsaved responses in localStorage (coming soon)

This additional cleans up some eslint warnings


penny_university_frontend/src/components/chats/ChatDetail.spec.tsx
```diff
@@ -57,7 +57,7 @@ test('users can only create follow ups when authenticated', () => {
   expect(markdownButton).toBeTruthy()
 })
 
-test('users can create follow ups when unauthenticated', () => {
+test('users can start a follow up whithout authentication', () => {
```
nchouard, line 60: nit: *without ;)

Issue #302, opened 2020-07-17
Add sentry to React app to catch issues that happen in-browser.


JnBrymn:
This https://github.com/getsentry/sentry-javascript/tree/master/packages/browser

PR #344, opened 2020-08-28
#302 Sentry setup

coreyar:
# Description
Minimal sentry setup
- Initializes sentry when the app mounts
- Adds the user to the scope when the user is fetched
- Clears the scope when the user logs out

# Proof of Completeness
Test Sentry errors sent from the browser to slack
https://penny-university.slack.com/archives/CHCM2MFHU/p1598610954199700

### DIFF
--- a/penny_university_frontend/src/config.ts
+++ b/penny_university_frontend/src/config.ts
@@ -1,5 +1,5 @@
 const config = {
-  sentry: 'https://cfb89b4197334bfebcedeb8c44ffbf06:978d276675c448d2b9f2e1257584cbaf@o410294.ingest.sentry.io/5284085'
+  sentry: 'https://cfb89b4197334bfebcedeb8c44ffbf06:978d276675c448d2b9f2e1257584cbaf@o410294.ingest.sentry.io/5284085',
 }
 
 export default config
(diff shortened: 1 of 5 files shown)


penny_university_frontend/src/config.ts
```diff
@@ -0,0 +1,5 @@
+const config = {
+  sentry: 'https://cfb89b4197334bfebcedeb8c44ffbf06:978d276675c448d2b9f2e1257584cbaf@o410294.ingest.sentry.io/5284085',
```
JnBrymn, line 2: Does this mean that we're storing this in the clear in the javascript in people's browsers? Practically speaking, at this point it isn't going to matter much, but it would be nice to not store it in the clear if we can avoid it.

But maybe I'm wrong. `import` can't be used outside of a module, so I don't know how I would get this value from the console.

coreyar:
Any unhandled error will go to Sentry. The only additional configuration we may need is adding extra context to error reporting to make them easier to track down.

penny_university_frontend/src/config.ts
coreyar: We could mangle the bundle so that it is harder to get. We could also make the key more secure by including it as a secret in the build process.

The worst that happens is that someone starts to send spam errors.

coreyar:
@nchouard `NODE_ENV` is managed by the create-react-app scripts.
https://create-react-app.dev/docs/adding-custom-environment-variables/
```There is also a built-in environment variable called NODE_ENV. You can read it from process.env.NODE_ENV. When you run npm start, it is always equal to 'development', when you run npm test it is always equal to 'test', and when you run npm run build to make a production bundle, it is always equal to 'production'. You cannot override NODE_ENV manually. This prevents developers from accidentally deploying a slow development build to production.```

PR #346, opened 2020-08-30
bug-135-first-followup-crash

coreyar:
# Description
On a first followup the user won't exist in the normalized entity. This PR adds a fallback in the selector to return the authenticated user in this situation


coreyar:
@nchouard Whoops I branched off the wrong base. The diff has been cleaned up to clearly show what changes resolves the issue

nchouard:
Gotcha, looks good to me!

JnBrymn:
Thanks you for taking care of this quickly! ❤️

Issue #360, opened 2020-09-16
Create "would you like to chat" smoke screen test.


JnBrymn:
We're considering making a donut clone but instead of having dedicated channels we would make _any_ channel a donut channel where we periodically ask if they want to be paired with people for topic-specific conversations.

Rather than build out the entire experience we need to gauge interest by making the first part of funnel and seeing if people want to set up a chat. The first part of the funnel is to ask each channel if they would like to be involved in a chat and then check how many people click "yes".

# TODO
- Create agreed upon slack blocks
- Identify channels and schedule to send them out
- Create tracking of how many clicks happen.
- Do it! Then review it!
- If we hit "criteria" then flesh out donut clone issues. 
  - [Review this doc](https://docs.google.com/document/d/1_bQGuUBnzeEROj8v7kiS_jUojGgEO2rdjUor39VVqKc/edit)
  - modify the existing donut issue

JnBrymn:
Thanks Nick!

PR #361, opened 2020-09-17
Set topic channels and allow users to request matches

nchouard:
# Description
Addresses part of #360. This change introduces a slash command that allows channels to be set as topics, a management command to send out some "request to match" blocks, and keeps track of the users who have clicked that they want to match with others.

## Details
You can use `/penny set-topic` (up for suggestions on what to change this command to) to set the channel as a topic channel, so that users can be set up to make matches in that channel. 
Screen Shot 2020-09-16 at 8 39 00 PM

When there are topic channels, you can use this management command to send the request match blocks to all the topic channels: `python manage.py request_matches`. You can also pass arguments to select specific topic channels.
Screen Shot 2020-09-16 at 8 39 17 PM

If the user matches, we add a record to the matchrequest table, and send them a message confirming that we will set them up with a chat in the near future.
Screen Shot 2020-09-16 at 8 39 34 PM


matchmaking/models.py
```diff
@@ -0,0 +1,23 @@
+from django.db import models
+
+from users.models import SocialProfile
+from pennychat.models import PennyChat
+
+
+class TopicChannel(models.Model):
+    channel_id = models.CharField(max_length=20, unique=True)
+    slack_team_id = models.CharField(max_length=20)
+    name = models.TextField()
```
JnBrymn, line 10: is `name` the topic associated with the channel or just the actual name of the channel cached so that we don't have to look it up again?

JnBrymn (approved):
This is a good start, however I would like to understand the setup behavior a little better. I can see in code how it would work, but are we even going to be testing this behavior for now? I was thinking that _we_ would decide which channels get the `request_match_blocks`, so the `set-topic` command isn't even needed not.

If `set-topic` command isn't needed now, then let's stick it into another branch and land the rest of the code without it. If we do need `set-topic` now, then I need to better understand the experiment you have in mind because I don't see how it would work yet. I understand the user experience, I just don't know what would trigger people to use the slack command.

matchmaking/models.py
nchouard: Name of the channel so that we don't have to look it up every time.

bot/management/commands/request_matches.py
```diff
@@ -0,0 +1,46 @@
...
+
+
+class Command(BaseCommand):
+    help = """Send out onboarding questionnaire to set of users specified by a list of slack names. If no users
+    are specified then all users that don't have a row in the database will be contacted. There is a safety switch in
+    the script requiring you to start the send after it lists who will get the onboarding."""
+
+    def add_arguments(self, parser):
+        parser.add_argument(
+            '--slack-team',
```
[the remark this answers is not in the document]
nchouard: The default gets all Slack teams, so it actually makes me more sane in the short term to have no default ID, because then I don't have to know the IDs for my QA env for testing vs the production env id.

nchouard:
I really just added the set topic command as a way for us to be able to really easily add topic channels, because when you use the slash command, you get all of the information about the channel. The alternative is to create the management command so that it gets every channel from slack and then matches based on the names you pass along, and creates those as topic channels and then sends the messages. The former seemed easier to me, so that's what I did. We would need some way of adding topic channels from Slack at some point anyway. If you would prefer that we do it by passing params I can do that instead.
I'm not intending to have other users use the command right now. If they did though, it would be totally fine. Not really an experiment, just an easy way for us to add topic channels.

PR #364, opened 2020-09-23
added review_match_requests command

JnBrymn:
# what
Created `review_match_requests` which prints the users that requests to be matched and the topics that they would like to be matched with.

# why
This will help us figure out how to pair people together for our 🍩 🔪 feature


JnBrymn:
@nchouard feel free to land this after you accept if I'm not around.

nchouard (approved):
Boom! Love it.

PR #366, opened 2020-10-05
make get_organizer -> get_organizers (plural)

JnBrymn:
The new Penny Chat channel conversation flow introduced a situation where a single chat could have two owner participants. (If one person shared a chat and then a different person shared it.) This caused a sentry error https://sentry.io/organizations/penny-university/issues/1937265091/?project=5284085&referrer=slack It's a corner case, but rather than remove the first organizer I thought it was more reasonable to allow multiple people to organize a chat... I see this as something we'll need anyway at some point.

This fixes it so that rather than have a `penny_chat.get_organizer()` that returns a User object, there is a `penny_chat.get_organizers()` that returns a list (a QuerySet actually) of Users.


JnBrymn:
This was done not on "vacation time" but on "vacation from my vacation time" ;)

PR #368, opened 2020-10-15
adding date field to Match

JnBrymn:
# What
As per title

# Why
There are 3 dates that are interesting to track in the match-making flow 1 - when they request be matched, 2 - when they are matched with others, 3 - when they actually create a penny chat from the match.

We currently do 1 and 3. This change is to do number 2.


deadbender (approved):
Looks good by me. Only concern I have is that `auto_now_add` forces the field to not be editable, right now I can't see any scenarios where we would need to edit that data though.

nchouard (approved):
I had thought about this exact thing while making the match reminders, so thank you for making it happen!

PR #371, opened 2020-10-20
Post ephemeral with fallback in Slack commands

### LINKED ISSUE #370: /penny chat in a private channel works but sends failure message
### Steps to Repro
In private channel type `/penny chat` and see:

image

but also see the modal pop up.

This might be confusing to users and lead them to mistrust the product is going to work.

nchouard:
# Description
Closes #370, and also allows the help message to be sent if someone makes a command in a private channel.

## Details
Makes the commands an `elif` block and uses the handy-dandy `postEphemeral_with_fallback` (thanks @JnBrymn) to send the help message.


deadbender (approved):
Looks good! Hitting a lot of small but big fixes these days :smile:

This reminds me - it may be good to update the ephemeral message for /penny...maybe linking back to a help page on the website or something of that sort.

JnBrymn (approved):
Great work!

PR #375, opened 2020-10-22
Colin/integration test rework

KO4JZT:
This solution is *extremely* hacky and more time consuming, but it's definitely better than the old tests.

I added in documentation for those new to Cypress as well.

Notes:
- I tried moving the variables used to a .env file to make things easier to find and work across multiple tests. Sadly, cy.setCookie() breaks when you use an environment variable, not sure why but I couldn't get it working.

# todo
- [x] Add instructions about how to get cookies to docs. (maintainence_guide.md is probably a good place)
- [x] Add log-in information to PennySecrets.
- [ ] Get John or Nick to run through the "how to run locally" docs.


JnBrymn:
This looks like a hopeful approach. Do you think you can demo it on Tuesday for us?

On thing to consider is that the ultimate goal is to be able to stick this into a bash script and just run it. If we can do that, then we will be able to stick it in the GitHub action and replace the existing integration tests. In this regard I'm a little concerned about the cookie stuff. It feels like that could cause these tests to be brittle.

There are 3 files that appear to be boilerplate and not really have any content. Do we need them? Should we delete them.

deadbender:
> On thing to consider is that the ultimate goal is to be able to stick this into a bash script and just run it. If we can do that, then we will be able to stick it in the GitHub action and replace the existing integration tests.

I believe this would be possible - the one fear I have is that you'd be exposing sign in cookies publicly in the file but it should also be possible to use it in the repo's secrets and use that. 

It definitely seems brittle, but from the looks of it it's the only option that gets around their security measures. Luckily, if we can find out if the sign in cookies hold up over time, it could definitely be a possibility to run it headless in the action. Only change that would need to be made for that is there would need to be an existing "mock" server running in a dev workspace to use for these tests, not sure yet how I could make that happen in the test itself.

>  There are 3 files that appear to be boilerplate and not really have any content. Do we need them? Should we delete them.

plugins/index.js, commands.js, and commands/index.js *should* be able to be deleted without any harm, will push up a commit with that change shortly here.

> Do you think you can demo it on Tuesday for us?

Sure can! Will throw it up into the PennyHack doc for the agenda.

deadbender:
Correction - looks like those 3 files *can't* be deleted. Cypress instantly regenerates them as soon as it is opened.

deadbender:
One month later, this test does still work (surprisingly)! It took an extra minute or two to open up, which I assume translates to it refreshing the token, but it went through (until it failed because my bot wasn't online). @JnBrymn @nchouard

JnBrymn:
Well, so we do need easier integration tests. Let's talk about this at the meeting tonight. I think the remainder of this issue is:

* Have a documented plan for what to do if the auth cookie approach quits working because the cookie expires (which I imagine it will eventually).
* Make sure that the tests are in parity with existing ones. And add more or at least create an issue to add more.
* Set up a QA environment and stick all of this stuff in the GitHub actions.

deadbender:
Turning this into a draft while I finalize this this week. Sorry this is taking so long, I've been trying to finish up some big work projects + this KOA event user testing starts this week.

.github/workflows/test.yml
```diff
@@ -88,3 +86,38 @@ jobs:
...
+          COOKIE_DS=1608437298
+          COOKIE_D=iBbiv1a2Ixu14A7O7CeranbsfNmGX%2FcWQmzUAJSVrIASzxy6LJpkLVSzbeVrn4cztStibH7N7Ia5KAHpTvTmw%2FeoTol8LVMuJiYp7eE5ftiRdYZmIjJS8YAfJU3J9UAL%2BYbZZo3Hb0kGbq1aS2p8AOedMe7BMcp5zg78bQuHG7TNnCmvb9z2G1JowA%3D%3D
+          COOKIE_IAP2=treatment
+          COOKIE_OPTANONCONSENT=isIABGlobal=false&datestamp=Sat+Dec+19+2020+22%3A09%3A36+GMT-0600+(Central+Standard+Time)&version=6.10.0&hosts=&consentId=176e7236-c06f-4963-b9ec-2ff0628f3bf7&interactionCount=1&landingPath=NotLandingPage&groups=C0004%3A0%2CC0002%3A1%2CC0003%3A1%2CC0001%3A1&AwaitingReconsent=false
+          COOKIE_X=azmcw4cvyxph1qzhlbsxu8gu0.1608437123"
+      - name: Cypress Run
+        uses: cypress-io/github-action@v2
+        with:
+          headless: true
+          project: ./penny_university/tests/integration
```
JnBrymn, line 123: nit - add new line

penny_university/tests/integration/cypress/integration/penny_chat.spec.js
```diff
@@ -0,0 +1,61 @@
...
+describe('/penny chat', () => {
+    it('creates a new chat', () => {
+
+        cy.setCookie('b', COOKIE_B)
+        cy.setCookie('d-s', COOKIE_DS)
+        cy.setCookie('d', COOKIE_D)
+        cy.setCookie('ec', COOKIE_EC)
+        cy.setCookie('iap2', COOKIE_IAP2)
+        cy.setCookie('OptanonConsent', COOKIE_OPTANONCONSENT);
+        cy.setCookie('x', COOKIE_X)
```
JnBrymn, line 30: Is it possible to toss this into a function and then use that function here and below?

JnBrymn (approved):
hurray! let's get this landed tomorrow night

PR #378, opened 2020-11-15
Gcal Integration

nchouard:
Opening a draft PR for now. Includes the work that I demoed on Tuesday. Please comment with any concerns!


JnBrymn:
* Keep in mind any place where you _might_ be adding any more effort on the part of our users. You changed the `Share` to `Submit` for a bit but changed it back. In the final state I _think_ that we share the chat as soon as the modal closes. Is that right? I just want to make sure that we don't have a new, _required_ step at the end.
* If I choose not to connect google, I'm going to see the same modal pop up after every chat I create. (Right?) That's going to feel a bit intrusive. What do you think about just making that an ephemeral post?
* `get_or_create_social_profile_from_slack_id` was written a little sloppily in that it calls to slack upon every request. Maybe memoize the function so that it's used at most once per call.

bot/tasks/pennychat.py
```diff
@@ -197,11 +197,13 @@ def _penny_chat_details_blocks(penny_chat_invitation, mode=None):
     if include_calendar_link:
         start_date = penny_chat_invitation.date.astimezone(utc).strftime('%Y%m%dT%H%M%SZ')
         end_date = (penny_chat_invitation.date.astimezone(utc) + timedelta(hours=1)).strftime('%Y%m%dT%H%M%SZ')
+
+        description = f'{penny_chat_invitation.description} [Video Link]({penny_chat_invitation.video_conference_link})'
```
JnBrymn, line 230: maybe add a new line? `...description}\n[Video...`

bot/tasks/pennychat.py
```diff
@@ -240,6 +242,45 @@ def _penny_chat_details_blocks(penny_chat_invitation, mode=None):
         date_time_block
     ]
 
+    if penny_chat_invitation.video_conference_link:
```
JnBrymn, line 274: probably a nit pick, but rather than have 3 sections here, I'd just have a `{PREVIEW, INVITE, UPDATE}` section with a parenthetical `(A video link will be provided shortly before the chat starts)` and a `REMIND` section with the full details.

bot/tasks/pennychat.py
```diff
@@ -240,6 +242,45 @@ def _penny_chat_details_blocks(penny_chat_invitation, mode=None):
...
+        body.append(
+            {
+                'type': 'section',
+                'text': {
+                    'type': 'mrkdwn',
+                    'text': f'*Video Call Link*'
```
JnBrymn, line 280: Nit: If there is no formatting, then don't use an f-string. It probably doesn't matter much, but I bet there is an extra parsing step that happens. (below as well)

integrations/google.py
```diff
@@ -0,0 +1,114 @@
...
+from django.conf import settings
+
+
+class GoogleCalendar:
+    def __init__(self, credentials, calendar_id='primary'):
+        self.service = build('calendar', 'v3', credentials=credentials)
+        self.calendar_id = calendar_id
+
+    @property
+    def events(self):
```
JnBrymn, line 21: I'm curious, why not just create `self.events = self.service.events()` in the init? 🤔 I wonder what calling `events()` each time you want `self.events` does.

deadbender:
Agreed with John on everything here.

If we're moving to a .env here, it might also be good to do that for SLACK_API_KEY as well.

From a user experience perspective, we might want to include a short blurb on the connection modal detailing what data we access (from what I remember Google's OAuth page for this is very vague) and what benefit they get from connecting Google. ex. `When you connect with Google, we use your calendar to keep you updated on your upcoming Penny Chats and automatically create video chats for you!`

Really like the way you structured this in the code, makes it simple to expand on for other integrations. I think the more an app integrates with other existing tools, the more a company or organization is willing to adopt it into that workflow, and starting with this we'll have a lot of room to grow on that idea.

integrations/views.py
```diff
@@ -1,37 +1,36 @@
...
-        # re-prompting the user for permission. Recommended for web server apps.
-        access_type='offline',
-        # Enable incremental authorization. Recommended as a best practice.
-        include_granted_scopes='true',
-    )
+def auth_success(request):
+    error = request.GET.get('error')
+    # TODO: Redirect to frontend page
+    if error:
+        return HttpResponse(f"<h1/>There was an error authorizing with Google.</h1><p>{request.GET.get('error')}</p>")
```
JnBrymn, line 13: We should probably log details to sentry here. Probably `sentry_sdk.capture_message` since this isn't a raised exception.

JnBrymn (commented):
I think it looks good so far. The main thing I'm wondering about is the user experience in the various paths through this. Would you mind demoing that again this week? (Especially if anything has changed.)

bot/tasks/pennychat.py
nchouard: Yeah flake caught this and I fixed in my most recent commit

bot/tasks/pennychat.py
nchouard: Lol, I totally forgot to revisit this. Adding to todos...

integrations/google.py
nchouard: Good point, I should do that instead.

bot/tasks/pennychat.py
nchouard: So just remove the **Video Link** header from the invite and add parentheses?

nchouard:
@JnBrymn 
> In the final state I think that we share the chat as soon as the modal closes. Is that right? I just want to make sure that we don't have a new, required step at the end.

Yes, this is correct! I didn't mean to originally commit the wording change on the button. I was messing around with having several modals at one point and forgot to change it back before committing.

> What do you think about just making that an ephemeral post?

I actually moved to this in my latest commit before seeing your message! Great minds, yada yada...

@deadbender 
> If we're moving to a .env here, it might also be good to do that for SLACK_API_KEY as well.

My intention isn't for everyone to move to a `.env`, I just did so because I prefer it to `direnv` which I was using before. Contributors can use whatever they like for setting up their environments which is why it is gitignored. We don't want to commit those files with our API keys to the repo.

> From a user experience perspective, we might want to include a short blurb on the connection modal detailing what data we access (from what I remember Google's OAuth page for this is very vague) and what benefit they get from connecting Google.

The message from Google says that it will allow us to "view and edit events" on their calendar. I think that is clear enough. I will definitely add that language you suggested for the benefits though!

bot/processors/pennychat.py
```diff
@@ -9,7 +9,7 @@
     post_organizer_edit_after_share_blocks,
     share_penny_chat_invitation,
     add_google_meet,
-    add_google_integration_blocks,
+    add_google_integration_blocks, update_google_meet,
```
JnBrymn, line 12: nit, new line

bot/tasks/pennychat.py
```diff
@@ -67,37 +66,57 @@ def post_organizer_edit_after_share_blocks(penny_chat_view_id):
...
+    penny_chat_invitation.video_conference_link = meet.get('hangoutLink')
+    penny_chat_invitation.google_event_id = meet.get('id')
     penny_chat_invitation.save()
 
 
+@background
+def update_google_meet(penny_chat_id):
+    penny_chat_invitation = PennyChatSlackInvitation.objects.get(id=penny_chat_id)
+
+    calendar = get_user_google_calendar_from_slack_id(penny_chat_invitation.organizer_slack_id)
```
JnBrymn, line 110: if the user retracts google permissions then calendar will be None here, add an `if calendar is None` here or we'll error on line 112.  In this case I guess we just keep the calendar invite as is? (I presume it will still work)

bot/tasks/pennychat.py
JnBrymn: yup, just like you did - looks good

integrations/signals.py
```diff
@@ -0,0 +1,18 @@
...
+from pennychat.models import PennyChatSlackInvitation
+
+
+@receiver(post_save, sender=GoogleCredentials)
+def add_google_meet_to_upcoming_chats(sender, **kwargs):
+    credentials = kwargs.get('instance')
+    if credentials:
+        user = credentials.user
+        slack_ids = [profile.slack_id for profile in user.social_profiles.all()]
+        invites = PennyChatSlackInvitation.objects.filter(organizer_slack_id__in=slack_ids, date__gt=timezone.now())
```
JnBrymn, line 16: would there ever be any invites with `date__gt=timezone.now()` ? is this a bug

JnBrymn: disregard ... hey this box wine is great!

bot/tasks/pennychat.py
```diff
@@ -197,11 +279,13 @@ def _penny_chat_details_blocks(penny_chat_invitation, mode=None):
...
+
+        description = f'{penny_chat_invitation.description}\nVideo Link: {penny_chat_invitation.video_conference_link}'
         google_cal_url = 'https://calendar.google.com/calendar/render?' \
                          'action=TEMPLATE&text=' \
-            f'{urllib.parse.quote(penny_chat_invitation.title)}&dates=' \
-            f'{start_date}/{end_date}&details=' \
-            f'{urllib.parse.quote(penny_chat_invitation.description)}'
+                         f'{urllib.parse.quote(penny_chat_invitation.title)}&dates=' \
+                         f'{start_date}/{end_date}&details=' \
+                         f'{urllib.parse.quote(description)}'
```
JnBrymn, line 288: this function has grown in complexity, can you pull out the contents of this if statement into it's own helpful func that lives in this file?

JnBrymn (approved):

nchouard:
Before merging todos:
- [ ] Add production google api and add to secrets
- [ ] Update docs

nchouard:
I have submitted to google for verification... And now we wait.

PR #379, opened 2020-11-16
Chats on profile page ordered incorrectly

### LINKED ISSUE #362: Penny Chats in wrong order on profile page
# Steps to Reproduce
~I _think_ this works:~ I tried this in my QA slack and couldn't reproduce the effect.
* Create 2 Penny Chats, Chat A for 10 minutes from now and Chat B for 20 minutes from now.
* Change Chat A to be 30 minutes from now.
* Load the profile page see that the Chat B is first and Chat A is next even though Chat A is further into the future and is even labeled as such. 

[This is definitely a front-end bug](https://github.com/penny-university/penny_university/issues/362#issuecomment-717681746)

Example of mis-ordered chats:
image

deadbender:
Fix pushed up from the other nights PennyHack for the ordering bug. Fixes #362 

Left redux-devtools in package.json since I think it could continue to be very useful.

Front-end tests should pass once I merge John's fix in his PR, didn't want to create any issues duplicating that.


penny_university_frontend/src/middleware/user.ts
```diff
@@ -36,7 +35,7 @@ const user : Middleware<Dispatch> = (store: MiddlewareAPI) => (next: (action: An
       break
     case Actions.BOOTSTRAP:
       checkAuth(store.dispatch)
-      store.dispatch(loadChatsList(ApiRoutes.chats))
+      // store.dispatch(loadChatsList(ApiRoutes.chats))
```
JnBrymn, line 38: go ahead and just delete this line

penny_university_frontend/src/middleware/user.ts
```diff
@@ -50,7 +49,7 @@ const user : Middleware<Dispatch> = (store: MiddlewareAPI) => (next: (action: An
     case Actions.FETCH_USER_SUCCESS:
       // Load data
       const pk = action.payload.result.pk.toString() // eslint-disable-line no-case-declarations
-      store.dispatch(loadChatsList(ApiRoutes.userChats(pk), pk))
+      // store.dispatch(loadChatsList(ApiRoutes.userChats(pk), pk))
```
JnBrymn, line 52: go ahead and just delete this line

JnBrymn (commented):
It's nice how the code is simpler. It feels like we need a test, but I don't know what that would be. Can we test that upon page load we only call the chats API once?

If possible, I'd like to get @nchouard or @coreyar to ✅ since they're more knowledgeable. 

Question: are we ok with keeping the dev-tool in the code like we have them? Do they automatically turn off prod? Should we put them in a separate "dev" file instead of package.json? (I'm just thinking about how our dev-requirements.txt works in Python.)

deadbender:
Not sure how to do a test just yet, but on the note of the dev-tool in package.json I went ahead and moved it to devDependencies - now it won't install if ENVIRONMENT=production or if you run it as `npm install --production`.

nchouard (commented):
What happened to the changes that were made to the Chats page (`pages/Chats.tsx`)? I thought that we had to change that in order to load chats properly on the main chats page. Can you add screenshots of the before/after of these changes?

penny_university_frontend/src/tests/Auth.test.ts
```diff
@@ -20,7 +19,7 @@ describe('auth flow', () => {
     document.cookie = ''
```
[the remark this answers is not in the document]
nchouard: Actually, this could be your test that we _don't_ fetch the chats list when we bootstrap.

deadbender:
Commit right before the first one of this PR

Latest commit in this PR

JnBrymn (approved):
"sounds good then" ~nick

Issue #380, opened 2020-11-19
Forgot password does nothing!


JnBrymn:
# steps to reproduce
* visit https://www.pennyuniversity.org/chats/ and log out
* click log-in
* enter your email address
* when prompted to enter your password, click "Forgot Password"
* see that nothing happens and http requests are even made!

the expectation here is that we would walk the users through a forgot password flow

PR #381, opened 2020-11-19
Fix broken password reset flow and test

JnBrymn:
# what
added back a missing parameter to `AuthPasswordModal`, added tests

# why
because it was breaking the forgotten password flow

# related
closes https://github.com/penny-university/penny_university/issues/380

### DIFF
--- a/penny_university_frontend/src/components/modal/auth/password.tsx
+++ b/penny_university_frontend/src/components/modal/auth/password.tsx
@@ -44,7 +44,9 @@ export const AuthPasswordModal = ({ email, login, requestPasswordReset }: AuthPa
         </Form>
       </ModalBody>
       <ModalFooter>
-        <Button id="forgot-password" color="link" onClick={() => { requestPasswordReset({ email }) }}>Forgot password?</Button>
+        <Button id="forgot-password" color="link" onClick={() => { requestPasswordReset({ email }) }}>
+          Forgot password?
+        </Button>
       </ModalFooter>
     </>
   )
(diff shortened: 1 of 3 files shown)


penny_university_frontend/src/components/modal/auth/password.tsx
```diff
@@ -17,7 +17,7 @@ type AuthPasswordModalProps = {
   requestPasswordReset: (payload: {email: string}) => void,
 }
 
-const AuthPasswordModal = ({ email, login }: AuthPasswordModalProps) => {
+export const AuthPasswordModal = ({ email, login, requestPasswordReset }: AuthPasswordModalProps) => {
```
nchouard, line 20: I don't think we need this export. In the tests, just import the default like `import AuthPasswordModal from '../../components/modal/auth/password'`. If that doesn't work, then this is fine.

penny_university_frontend/src/components/modal/auth/password.tsx
```diff
@@ -44,7 +44,7 @@ const AuthPasswordModal = ({ email, login }: AuthPasswordModalProps) => {
         </Form>
       </ModalBody>
       <ModalFooter>
-        <Button color="link" onClick={() => { requestPasswordReset({ email }) }}>Forgot password?</Button>
+        <Button id="forgot-password" color="link" onClick={() => { requestPasswordReset({ email }) }}>Forgot password?</Button>
```
nchouard, line 47: Run `npm run lint:fix` locally

nchouard (approved):

penny_university_frontend/src/components/modal/auth/password.tsx
JnBrymn: That's how it originally was and I had to change it to this. I noticed that `Alerts.test.tsx` was testing the exported `ErrorAlert` rather than the default exported `connect(...)(ErrorAlert)`.

PR #384, opened 2020-12-02
Fix chat ordering

deadbender:
This PR (in local testing) has proven to fix the issues caused through removing the BOOTSTRAP action grab of the chat data. I basically just copied over what we had on the profile page that is working, and it seems to work here and now creates only 1 request on /chats.


JnBrymn (approved):
LGTM

penny_university_frontend/src/pages/Chats.tsx
```diff
@@ -1,8 +1,36 @@
...
 
-const ChatsPage = () => (
-  <ChatList filter={{ key: 'all', query: '' }} />
-)
+type DispatchProps = {
+  loadChatsList: (nextPageUrl?: string, userID?: string) => void,
+}
 
-export default ChatsPage
+type TParams = { id: string };
```
nchouard, line 14: Why is this here?

penny_university_frontend/src/pages/Chats.tsx
```diff
@@ -1,8 +1,36 @@
...
+import { RouteComponentProps } from 'react-router-dom'
+import ChatList from '../components/chats/ChatList'
+import { loadChatsList } from '../actions/chat'
+import ApiRoutes from '../constants'
 
-const ChatsPage = () => (
-  <ChatList filter={{ key: 'all', query: '' }} />
-)
+type DispatchProps = {
+  loadChatsList: (nextPageUrl?: string, userID?: string) => void,
```
nchouard, line 11: Do we need userID here?

penny_university_frontend/src/pages/Chats.tsx
```diff
@@ -1,8 +1,36 @@
...
+  }, [loadChatsList])
+  return (
+    <div>
+      <ChatList filter={{ key: 'all', query: '' }} />
+    </div>
+  )
+}
+
+const mapDispatchToProps = (dispatch: ThunkDispatch<unknown, unknown, AnyAction>) => ({
+  loadChatsList: (nextPageUrl: string, userID: string) => dispatch(loadChatsList(nextPageUrl, userID)),
```
nchouard, line 32: Do we need to pass userID here?

nchouard (commented):
I think you have some unnecessary props/parameters you should remove

PR #386, opened 2020-12-04
Add privacy policy page

nchouard:
Added a privacy policy page at `/privacy`, and added links to the footers in our app.
Screen Shot 2020-12-01 at 9 12 22 PM
Screen Shot 2020-12-01 at 9 16 06 PM
Screen Shot 2020-12-03 at 9 21 04 PM

### DIFF
--- a/home/templates/home/base.html
+++ b/home/templates/home/base.html
@@ -39,7 +39,55 @@
 
 <body>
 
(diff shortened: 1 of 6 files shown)


JnBrymn (approved):
image

Issue #385, opened 2020-12-02
Fix npm audit vulnerabilities on frontend


KO4JZT:
When running `npm install`, it reports back about 5000 low vulnerabilities. `npm audit fix` solves most of these, except some issues with getting sass-lint and react-scripts packages fixed, and introduces some new issues - specifically issues with eslint not wanting to work correctly with @typescript-eslint/eslint-plugin and its no-redeclare rule. eslint 7.11.0 also seems to be a lot stricter, with no-unused-vars causing react-scripts to not build the application.

PR #398, opened 2021-01-24
Front-end Vulnerability Fixes

KO4JZT:
Fixes all vulnerabilities from #385 except 3 (sass-lint-auto-fix does not have an update, but the vulnerabilities are low severity in one package so I assume it is fine). Tested and React App does still work exactly as before - all tests succeed 👍🏻.


KO4JZT:
Hm, not sure why Eslint crashed after 2sec. On mobile now but will take a look on my computer in an hour or two.

Tests and linking all passed here - maybe workflow needs some updating?

JnBrymn (approved):
Thanks for keeping us safe!

nchouard (approved):
Looks good! Thank you for taking care of something important but too often ignored.

